PR #1, opened 2016-12-19
Exclude gtest if LAF_WITH_TEST is disabled

winterheart:
Save some time during compilation and exclude unnecessary test framework when LAF_WITH_TEST=OFF


dacap:
Hi @winterheart, thanks for this change. I think this can be useful [for people running `make install`](https://github.com/aseprite/aseprite/issues/1366) (so gtest is not installed). Thanks!

PR #6, opened 2019-03-20
Fix Caps Lock has no effect on macOS

kawa-yoiko:
Hi! Hope this can be of a little help (^ ^;)

It is a not-so-elegant workaround — I cannot find much information on how `UCKeyTranslate()` works, and this is the only way I can get the outcome correct so far. For now it seems that `UCKeyTranslate()` ignores the Caps Lock flag without Shift, and reads it with Shift, hence always disabling its function. I don't know the reason behind it, and I might, might be terribly wrong… I'd appreciate if anyone can provide a deeper insight into it!

Btw, this can fix aseprite/aseprite#1876.

### DIFF
--- a/os/osx/keys.mm
+++ b/os/osx/keys.mm
@@ -409,13 +409,21 @@ CFStringRef get_unicode_from_key_code(const UInt16 keyCode,
   UniChar output[4];
   UniCharCount length;
 
+  // As UCKeyTranslate doesn't care about the Caps Lock flag,
+  // translate it into Shift
+  NSEventModifierFlags modifierFlagsHacked = modifierFlags;
+  if (modifierFlagsHacked & NSEventModifierFlagCapsLock) {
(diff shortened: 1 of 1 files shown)


dacap:
Hi @kawa-yoiko! Thanks for your help on this one. As you said this looks a little hacky, the main issue with this patch is that we could introduce a new problem: Caps lock will act like Shift, so numbers will write symbols (`1 -> !`, `2 -> @`, etc.) which is not desired.

I've tried to find an alternative solution and Chromium code was really helpful. You can find the fix here 75c2c3c5e98f7de2248a704f8e9fe5cd78722474. Thanks anyway for your PR!

kawa-yoiko:
Ah indeed, I did not notice the new problem and did not know such an orderly approach existed. Thanks for your effort ^ ^

PR #7, opened 2019-03-21
Fix incorrect behaviour of macOS native dialogs

kawa-yoiko:
Please refer to the commits for details. I have tested this on macOS 10.13.5, and everything about native dialogs and menus look fine now. (Refer to aseprite/aseprite#2037.)

On a sided note, I cannot reproduce aseprite/aseprite#1917, but it seems a related issue and I suppose it should be looked into.

Thanks for the effort and time! ^ ^

### DIFF
--- a/os/osx/native_dialogs.mm
+++ b/os/osx/native_dialogs.mm
@@ -68,6 +68,8 @@
   }
 
   display->setNativeMouseCursor(oldCursor);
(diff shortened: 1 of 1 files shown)


dacap:
Thanks for this one @kawa-yoiko. Just merged!

I have updated the instructions to reproduce the aseprite/aseprite#1917 bug and I can confirm that this PR fixes that bug and others 👍

PR #8, opened 2019-03-28
Extend rectangle packing algorithm with padding support

kawa-yoiko:
I included a simple test, which has already helped me catch a bug ^ ^

Combined with aseprite/aseprite#2045, this tries to address aseprite/aseprite#1240.

player

Hope it helps!


dacap:
Thanks again @kawa-yoiko! Just merged

PR #9, opened 2019-06-20
Keep supplied extension in macOS save dialogues

kawa-yoiko:
The dialogue should keep the extension if it's among the allowed ones, or append the first one in the allowed list to the file name otherwise. Closes aseprite/aseprite#1835.

For some reason `setAllowedFileTypes:` overwrites the extension provided in `setNameFieldStringValue:`, so it needs to be done before the latter. Apparently [authors of electron have run into the same issue](https://github.com/electron/electron/blob/34c4c8d5088fa183f56baea28809de6f2a427e02/shell/browser/ui/file_dialog_mac.mm#L179)... Does this count as an upstream bug in AppKit? (=~=)

### DIFF
--- a/os/osx/native_dialogs.mm
+++ b/os/osx/native_dialogs.mm
@@ -118,13 +118,6 @@ public:
       [panel setTitle:[NSString stringWithUTF8String:m_title.c_str()]];
       [panel setCanCreateDirectories:YES];
 
-      std::string defPath = base::get_file_path(m_filename);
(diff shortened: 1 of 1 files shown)


dacap:
Hi @kawa-yoiko, I've tested this change on macOS 10.14.5 and the file extension is still not showing up 😢 Anyway after applying an extra change (c1a4379bf215caeb934f299bd684c8f2c0680e3c) it does work, so I've merged this change 👍

PR #10, opened 2019-06-20
Fix macOS dialogues not responding to Cmd-AZXCV

kawa-yoiko:
Closes aseprite/aseprite#1973.

I was sure this could not be achieved with an application delegate; an invisible menu would solve the problem, but might be quite cumbersome. Thus I ended up copying much code in [this StackOverflow post](https://stackoverflow.com/a/3176930) and it worked nicely.

I'm not sure whether this is the right way to name ("OSX" + "NS" looks unnecessary, but there is already an "OSXApp") and place the inherited class, so please specify if anything is to be changed. Thanks!

### DIFF
--- a/os/osx/app.mm
+++ b/os/osx/app.mm
@@ -21,7 +21,7 @@ namespace os {
 class OSXApp::Impl {
 public:
   bool init() {
-    m_app = [NSApplication sharedApplication];
+    m_app = [OSXNSApplication sharedApplication];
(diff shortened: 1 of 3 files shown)


dacap:
Thanks for this one too @kawa-yoiko 🙏 (`OSXNSApplication` is fine, maybe in the future we rename it to `OSXApplication` or similar)

PR #13, opened 2019-08-13
Include path fix for newest version of Skia

dpacbach:
Latest version of skia used by aseprite has a change to folder structure which is causing SKIA_SKCMS_INCLUDE_DIR to be NOTFOUND which is breaking the aseprite build.


dacap:
Hi @dpacbach, actually there are several changes in newest Skia versions, we're working on a refactor of the library in the [refactors](https://github.com/aseprite/laf/tree/refactors) branch (which will start using Skia m76).

PR #16, opened 2020-05-28
Add convert_to_4_digits_string function

Gasparoken:
This addition makes possible to fix issue #2388 Recent Files history order at the home view.
When this commit has been merged to master, next step is to merge the PR of the branch Gasparoken/fix-recent-files to finally solve issue #2388.


dacap:
@Gasparoken sorry maybe I wasn't clear enough in the email, I think that we should not include this function in laf because it's not generic enough. We can just use `fmt::format()` for these kind of formatting.

PR #19, opened 2020-07-04
[osx] Use Retina display scale

nanoant:
On Retina displays view/window coordinates are logical points not pixels and `window.backingScaleFactor` represents how many pixels are covered by single coordinate point.

This change effectively makes LAF UI 2x smaller on Retina displays, but allows drawing at native Retina resolution.

A picture is worth a thousand words: (NOTE: Click on the image to see it on full 2x/Retina scale)
image

Picture above depicts Aseprite using 400% scale on Retina and using TTF font instead of bitmap font.

I understand this change may be completely irrelevant for most of the Aseprite users using default UI settings, but if you want to use TTF font this change actually makes all TTF fonts look crisp on macOS Retina displays. And I prefer to have TTF font on my Retina display rather than default bitmap font, that I find hard to read 🙄 

### DIFF
--- a/os/osx/view.mm
+++ b/os/osx/view.mm
@@ -50,8 +50,9 @@ gfx::Point get_local_mouse_pos(NSView* view, NSEvent* event)
     scale = [(OSXWindow*)[view window] scale];
 
   // "os" layer coordinates expect (X,Y) origin at the top-left corner.
-  return gfx::Point(point.x / scale,
-                    (view.bounds.size.height - point.y) / scale);
+  float displayScale = view.window.backingScaleFactor;
+  return gfx::Point(point.x / scale * displayScale,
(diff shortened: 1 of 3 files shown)


dacap:
Hi @nanoant, I like this because it makes even more similar to the Windows implementation (which uses the high DPI resolution). I'll review this next week (still not sure if we should make this optional). Thanks for the contribution!

nanoant:
@dacap Great, thanks. Indeed this makes the macOS Aseprite behavior on par with Windows. I was first using Aseprite on my Windows PC machine for Godot development, and then when I tried to use it on my Mac, I found out that despite being `NSHighResolutionCapable` Aseprite is not drawing at full Retina res. Hence the PR.

FYI I've just updated my PR as I forgot to handle display scale `OSXWindow::setScale`. I hope now all the necessary code paths are covered.

PR #21, opened 2020-08-11
Fix hue set function.

Gasparoken:
Before this fix the hue set function clamped the hue value between 0 and 360, instead of convert the hue value to an equivalent angle between 0 and 360.


dacap:
Merged with a [minor fix](https://github.com/aseprite/laf/commit/8b17e61e6a6c29a6c64c784ed52a1ba057adef4b) to fix CI on MSVC:
https://ci.appveyor.com/project/dacap/laf/builds/34599173/job/3pl0p9tp2po9pqt2#L323

PR #27, opened 2021-06-03
X11 error when glibc bounds checks enabled (beta branch)

clarfonthey:
See aseprite/aseprite#2755.

### DIFF
--- a/os/x11/xinput.h
+++ b/os/x11/xinput.h
@@ -159,7 +159,7 @@ public:
 
     ASSERT(XSelectExtensionEvent);
     XSelectExtensionEvent(display, window,
-                          &m_eventClasses[0],
+                          m_eventClasses.data(),
                           int(m_eventClasses.size()));
   }
 


os/x11/xinput.h
```diff
@@ -159,7 +159,7 @@ class XInput {
 
     ASSERT(XSelectExtensionEvent);
     XSelectExtensionEvent(display, window,
-                          &m_eventClasses[0],
+                          m_eventClasses->data(),
```
dacap, line 162: m_eventClasses is not a pointer but a std::vector, so we have to access data() member with the . operator:
```suggestion
                          m_eventClasses.data(),
```

dacap (changes requested):
Hi @clarfonthey, thanks for this change, anyway it doesn't compile because m_eventClasses is not a pointer. Would you mind to patch this in `main` branch? (so we can update both versions, the main and beta branches)

clarfonthey:
Made change and filed #28 for main branch. Thanks! :)

PR #32, opened 2021-08-24
Simplify to_utf8 and from_utf8 functions

martincapello:
This PR replaces the use of `WideCharToMultiByte` and `MultiByteToWideChar` to convert to and from utf8 strings by the use of the C++11 standard conversion functions.

One thing to note is that the int parameter of `to_utf8` was removed, which breaks the compatibility with previous versions of the library. Not sure if we should keep it and adapt the function to work as before.

WARNING: C++17 deprecates the codecvt header...so...maybe this PR doesn't worth it.


dacap:
Hi @martincapello 👋  it looks like there are some problems with gcc on Ubuntu with this ([check the CI](https://github.com/aseprite/laf/pull/32/checks?check_run_id=3413757974)). I'm not sure but it looks like `std::codecvt_utf8` was deprecated on C++17. I'm not sure about the availability or performance of the standard conversion vs the Win32 impl. I've found this too:

https://stackoverflow.com/a/42946556/408239

martincapello:
I saw this is failing for linux, and I think I can make it work, but I need to know first if we really want to merge this, otherwise it would be a waste of time.

dacap:
From what I can see, it looks like codecvt_utf8 will be removed in the future:

* https://twitter.com/StephanTLavavej/status/837824211259289600
* https://twitter.com/janwilmans/status/1123983286701834241

But if you find other information we can see what to do.

martincapello:
> Hi @martincapello 👋 it looks like there are some problems with gcc on Ubuntu with this ([check the CI](https://github.com/aseprite/laf/pull/32/checks?check_run_id=3413757974)). I'm not sure but it looks like `std::codecvt_utf8` was deprecated on C++17. I'm not sure about the availability or performance of the standard conversion vs the Win32 impl. I've found this too:
> 
> https://stackoverflow.com/a/42946556/408239

Just saw that stackoverflow post a minute before your comment...so frustrating...will close this. Thanks.

dacap:
Anyway it was a nice try to remove that Windows specific code, and I didn't know about codecvt, maybe in the future (I'm not sure if in C++23) there will be some standard Unicode conversion functions.

PR #33, opened 2021-10-05
Add Sha1 class methods digest() and size()

martincapello:
Adds the following methods to the Sha1 class:
- digest(): returns a pointer to the digest data.
- size(): returns the digest size.


dacap:
It's strange, I've just pushed the same commit to `main` but this PR is not marked as merged. Closing it (but it's merged).

PR #35, opened 2021-11-01
Force zero init of X11 class.

slembcke:
Running Aseprite batch mode (using `-b`) as part of a headless asset build process was causing it to segfault in `XInput::~XInput()` because `m_xi` was uninitialized. Forcing zero initialization fixes it.

### DIFF
--- a/os/common/main.cpp
+++ b/os/common/main.cpp
@@ -62,7 +62,7 @@ int main(int argc, char* argv[]) {
   if (!app.init())
     return 1;
 #elif LAF_LINUX
-  os::X11 x11;
+  os::X11 x11();
 #endif
   return app_main(argc, argv);
 }


iamOgunyinka:
Isn't this what's known as most-vexing parse?

dacap:
Hi @slembcke, I think your change, as @iamOgunyinka said, might be converting the calling of `os::X11` constructor to create `x11` instance, to a function definition (so the instance is not created, and maybe the ctor/dtor not even called).

I've pushed a possible fix [here](https://github.com/aseprite/laf/commit/3ad116e5c34a5525b686f4e006feff24483760e5). Initializing the m_xi pointer (the pointer inside `XInput` class) correctly. Could you check if this fix your problem?

slembcke:
Oh no. >_< This is why friends don't let friends use C++. :p

Joking aside, that commit seems to fix the issue as well. Thanks.

PR #36, opened 2021-11-16
Fix crash on exit when color space is different from default system's

iamOgunyinka:
This PR fixes the situation where there are lingering events in the global `g_event` event queue that references already nullified window and thus trigger an ASSERT in the destructor of `WindowWin`. This usually happen when the color space is different from the system's default. The severity and technicality of this bug is discussed in the Aseprite repository [here](https://github.com/aseprite/aseprite/issues/3055).

### DIFF
--- a/os/common/system.h
+++ b/os/common/system.h
@@ -37,7 +37,10 @@ namespace os {
 class CommonSystem : public System {
 public:
   CommonSystem() { }
-  ~CommonSystem() { set_instance(nullptr); }
+  ~CommonSystem() {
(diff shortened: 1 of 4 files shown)


dacap:
Thanks Joshua, merged in https://github.com/aseprite/laf/commit/18f571bf9a1f3034c99040a0fd9855ca33d54b7e followed by https://github.com/aseprite/laf/commit/6eefcb859bfbf00a3d993d7bac6a44af93e00bd9

PR #37, opened 2021-12-16
Add copy_file function and its windows implementation

martincapello:
Adds a new copy_file function. This PR adds only the Windows based implementation. Linux and macOS implementations throw exception.


dacap:
Just merged 👍  I'll try to implement this for the other platforms next week.

PR #40, opened 2022-01-01
base::copy_file implementation

doomed-phobos:
Hi!

This change, in short, implements base::copy_file function based on the idea of copying and pasting memory blocks.
I've also changed the functions (move, copy and delete file) that throw exceptions unnecessarily to return bool type.
(It needs clarify that I'm not an expert :sweat_smile:)

### DIFF
--- a/base/fs.h
+++ b/base/fs.h
@@ -25,9 +25,10 @@ namespace base {
 
   size_t file_size(const std::string& path);
 
-  void move_file(const std::string& src, const std::string& dst);
-  void copy_file(const std::string& src, const std::string& dst, bool overwrite);
-  void delete_file(const std::string& path);
+  bool move_file(const std::string& src, const std::string& dst);
+  bool copy_file(const std::string& src, const std::string& dst, bool overwrite);
+  // TODO: Change bool to enum Result?
+  bool delete_file(const std::string& path);
 
   bool has_readonly_attr(const std::string& path);
   void remove_readonly_attr(const std::string& path);
--- a/base/fs_unix.h
+++ b/base/fs_unix.h
@@ -25,6 +25,7 @@
 
 #include "base/paths.h"
 #include "base/time.h"
+#include "base/file_handle.h"
 
 #define MAXPATHLEN 1024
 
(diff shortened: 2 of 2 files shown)


dacap:
Thanks for your contributino @doomed-phobos, at the moment we'd prefer to keep the original behavior of the functions (throwing exceptions).

I'll add some comments to your copy_file implementation just in case if you want to complete it.

base/fs.h
```diff
@@ -25,9 +25,10 @@ namespace base {
 
   size_t file_size(const std::string& path);
 
-  void move_file(const std::string& src, const std::string& dst);
-  void copy_file(const std::string& src, const std::string& dst, bool overwrite);
-  void delete_file(const std::string& path);
+  bool move_file(const std::string& src, const std::string& dst);
+  bool copy_file(const std::string& src, const std::string& dst, bool overwrite);
+  // TODO: Change bool to enum Result?
+  bool delete_file(const std::string& path);
```
dacap, line 31: Don't change the signature of these functions.

base/fs_unix.h
```diff
@@ -57,25 +58,46 @@ size_t file_size(const std::string& path)
...
-    throw std::runtime_error("Error moving file");
+  return std::rename(src.c_str(), dst.c_str()) == 0;
 }
 
-void copy_file(const std::string& src, const std::string& dst, bool overwrite)
+// Don't secure for interrumptions, write in a temp?
+bool copy_file(const std::string& src, const std::string& dst, bool overwrite)
 {
-  throw std::runtime_error("Error copying file: unimplemented");
+  if(compare_filenames(src, dst) == 0) // Those are same?
```
dacap, line 69: Use the coding style/format (space between `if (`):
```
if (compare_filenames(src, dst) == 0) // Those are same?
```

base/fs_unix.h
```diff
@@ -57,25 +58,46 @@ size_t file_size(const std::string& path)
...
+    return false;
+
+  if(is_file(dst) && !overwrite) // If dst exists, we need overwriting
+    return false;
```
dacap, line 73: Throw an exception that the file exists.

base/fs_unix.h
```diff
@@ -57,25 +58,46 @@ size_t file_size(const std::string& path)
...
+
+  auto src_file = open_file(src, "rb"), dst_file = open_file(dst, "wb");
```
dacap, line 75: This can be in two lines (just to make it clear that we are opening two files):
```c++
auto src_file = open_file(src, "rb");
auto dst_file = open_file(dst, "wb");
```

base/fs_unix.h
```diff
@@ -57,25 +58,46 @@ size_t file_size(const std::string& path)
...
+  if(!src_file || !dst_file) // both exists?
+    return false;
+
+  // Getting optimal block size for I/O
+  struct stat s_stat;
+  if(fstat(fileno(src_file.get()), &s_stat) != 0)
+    return false;
+
+  std::vector<char> block(s_stat.st_blksize);
```
dacap, line 84: I think it will be easier to just use a fixed size (e.g. 64k), instead of using fstat, but it would be nice to benchmark what is faster in these cases. We're doing exactly this in the read_file_content function: https://github.com/aseprite/laf/blob/50409e16d48220cc4bbb994c81ba7f0107b51350/base/file_content.cpp#L26

dacap (changes requested):

dacap:
Implemented with https://github.com/aseprite/laf/commit/f5ed6fc4dc629e152ff3de42acedfc1c5e601c4b

PR #43, opened 2022-03-05
Error messages in fs_unix.h

Delta2Force:

I have simply added `strerror` to each function which throws an error to get a more detailed error message.


dacap:
Thanks @Delta2Force, I'll merge this with some little correction e.g. keep using std::rename() as it is (or should be) defined in <cstdlib> inside the std namespace.

PR #44, opened 2022-07-13
Add Rect floor function

Gasparoken:
This 'floor' function is useful to adjust x, y float coordinates to the _left most near integer_ for x, and _top most near integer_ for y (instead of rounding towards 0).
This function was included to solve issue aseprite/aseprite#2891

Example of use:
```
gfx::RectF boundsF(-0.25, -0.75, 1, 2);
gfx::Rect bounds = boundsF.floor();
bounds = (-1, -1, 1, 2)
```
When 'floor' is not used:
```
gfx::RectF boundsF(-0.25, -0.75, 1, 2);
gfx::Rect bounds = boundsF;
bounds = (0, 0, 1, 2)
```

### DIFF
--- a/gfx/rect.h
+++ b/gfx/rect.h
@@ -1,5 +1,5 @@
 // LAF Gfx Library
-// Copyright (C) 2019-2021  Igara Studio S.A.
+// Copyright (C) 2019-2022  Igara Studio S.A.
 // Copyright (C) 2001-2017  David Capello
 //
 // This file is released under the terms of the MIT license.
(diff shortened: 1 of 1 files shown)


dacap:
Should this floor the size of the rectangle too? or should we call the function floorOrigin()?

Gasparoken:
> Should this floor the size of the rectangle too? or should we call the function floorOrigin()?

Since Rect size (w, h) are greater than 0, I think isn't need to apply floor on size. Yes, we can call floorOrigin() if you like.

dacap:
Oh I see, we can keep floor() in that case, I'll merge this ASAP 👍

dacap:
Merged in https://github.com/aseprite/laf/commit/871adbe48bc85cff8de08d243df1f42fc9f1f7e9 with a test.

PR #45, opened 2022-07-15
Add optional 'exception throw' to remove_directory function

Gasparoken:


dacap:
Hi @Gasparoken, please catch the exception in the caller point. There are plans to remove exceptions from these functions anyway but it's something for the future (as it needs a general refactor and some extra thinking).

PR #46, opened 2022-08-01
add PowerPC

lifning:


dacap:
Hi @lifning, thanks for your contribution. I saw that the commit doesn't contain a valid email address (which means a way to contact you for any future purpose). I'll close this PR but you can reopen it if the commit is fixed.

PR #48, opened 2022-10-15
Fix Windows build

chaosink:
```
find_library() is not a proper way for finding OpenGL libraries.
On Windows, SKIA_OPENGL_LIBRARY got "SKIA_OPENGL_LIBRARY-NOTFOUND"
without using VS-Dev-CMD.

See discussions here:
    https://gitlab.kitware.com/cmake/cmake/-/issues/19804
    https://gitlab.kitware.com/cmake/cmake/-/issues/20226

Use find_package(OpenGL) and OpenGL::GL instead.
```
```
On Windows systems with non-English locales, the wide strings may cause
"Error C2001: Newline in constant".

2 solutions:
    Save the file with UTF-8 with BOM (byte-order mark).
    Specify UTF-8 charset for compilation. Used by this commit.

See discussions here:
    https://github.com/libusb/libusb/issues/207#issuecomment-288664124
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+# CMake output
+/build/
--- a/cmake/FindSkia.cmake
+++ b/cmake/FindSkia.cmake
@@ -28,13 +28,8 @@ endif()
 
 # Skia library
 find_library(SKIA_LIBRARY skia PATH "${SKIA_LIBRARY_DIR}")
-if(WIN32)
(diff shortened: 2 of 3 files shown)


asasini11 (approved):

dacap:
Thanks @chaosink for this but I'm not sure that these changes are really required.

PR #49, opened 2022-11-12
Don't use pthread_detach() after pthread_join()

winterheart:
After pthread_join() all allocated to thread resources are freed, so pthread_detach() after pthread_join() will results in undefined behavior with SIGSERV on some libc implementations (like MUSL). According to pthread_detach(3), "Either pthread_join(3) or pthread_detach() should be called for each thread that an application creates".

### DIFF
--- a/base/thread.cpp
+++ b/base/thread.cpp
@@ -55,8 +55,12 @@ thread::thread()
 
 thread::~thread()
 {
-  if (joinable())
+  if (joinable()) {
+#if LAF_WINDOWS
+    ::CloseHandle(m_native_handle);
+#endif  
     detach();
+  }
 }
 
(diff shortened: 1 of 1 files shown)


dacap:
Thanks @winterheart, I'll take a look to this next week to merge it. There are plans to completely replace base::thread with std::thread in a near future (base::thread were pre-C++11 times).

dacap:
@winterheart thanks for your PR here, anyway we have just [removed base::thread and use std::thread instead](https://github.com/aseprite/laf/commit/4c588f96774b6d651ae3d6ca70e97f7e290011af), which was something planned from a long time ago. I guess that will be the best option to fix any issue in the old base::thread impl.

PR #55, opened 2023-02-21
Fix compilation on GCC13

winterheart:
GCC-13 changes internal cstdint includes, and now files that uses standard integer types should directly include cstdint header.

See: https://gcc.gnu.org/gcc-13/porting_to.html#header-dep-changes
Bug: https://bugs.gentoo.org/865117
Bug: https://bugs.gentoo.org/895616

### DIFF
--- a/base/file_content.h
+++ b/base/file_content.h
@@ -9,6 +9,7 @@
 #pragma once
 
 #include "base/buffer.h"
+#include "base/ints.h"
 
 #include <cstdio>
 #include <string>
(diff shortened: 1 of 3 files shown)


mochaaP (approved):
lgtm

dacap:
Just cherry-picked in 0673b7c5928317f409c15f1654f5ee3b82011ba5 thanks for the patch @winterheart and thanks @mochaaP for reviewing it 👍

PR #58, opened 2023-03-27
Fix selecting a frame on the timeline (with left click), and immediately right clicking it, will be taken as double click (fix #3713)

Gasparoken:
fix aseprite/aseprite#3713


Gasparoken:
Ok, I understood. So maybe we need the current and previous click to evaluate a real double click action.

Gasparoken:
Ready for review.

PR #63, opened 2023-06-28
build: fix linking against shared library

mochaaP:
When building with shared libraries is enabled in the top-level project (e.g. aseprite here), laf still tries to find the vendored source.
This PR fixes the issue by checking whether a library is specified to use shared libs.

---

Signed-off-by: Zephyr Lykos \<self@mochaa.ws>

### DIFF
--- a/cmake/FindSkia.cmake
+++ b/cmake/FindSkia.cmake
@@ -128,14 +128,22 @@ if(NOT ZLIB_LIBRARIES)
   set(ZLIB_LIBRARIES ${ZLIB_LIBRARY})
 endif()
 
+if(USE_SHARED_FREETYPE)
+find_library(FREETYPE_LIBRARY freetype2)
+else()
 set(FREETYPE_FOUND ON)
 find_library(FREETYPE_LIBRARY freetype2 PATH "${SKIA_LIBRARY_DIR}" NO_DEFAULT_PATH)
 set(FREETYPE_LIBRARIES ${FREETYPE_LIBRARY})
 set(FREETYPE_INCLUDE_DIRS "${SKIA_DIR}/third_party/externals/freetype/include")
+endif()
 
+if(USE_SHARED_HARFBUZZ)
+find_library(HARFBUZZ_LIBRARY harfbuzz)
+else()
 find_library(HARFBUZZ_LIBRARY harfbuzz PATH "${SKIA_LIBRARY_DIR}" NO_DEFAULT_PATH)
 set(HARFBUZZ_LIBRARIES ${HARFBUZZ_LIBRARY})
 set(HARFBUZZ_INCLUDE_DIRS "${SKIA_DIR}/third_party/externals/harfbuzz/src")
+endif()
 
 set(SKIA_LIBRARIES
   ${SKIA_LIBRARY}


dacap:
Hi @mochaaP, actually USE_SHARED_FREETYPE and USE_SHARED_HARFBUZZ are flags specified for Aseprite (not for laf). In this case, if we use Skia as a backend, I think we should check the args.gn file of the Skia compilation to check if we can link against the static version of these libraries or use the system ones ([these flags](https://github.com/aseprite/skia#skia-on-linux) are specified when Skia is built with `skia_use_system_freetype2=false` and `skia_use_system_harfbuzz=false`). 

E.g. something like this is done to check if laf is being compiled with the same flags as Skia [here](https://github.com/aseprite/laf/blob/0673b7c5928317f409c15f1654f5ee3b82011ba5/cmake/FindSkia.cmake#L42).

mochaaP:
Surely. I suppose we should also ensure that these two flags are in sync?

---

by the way, it seems static linking is broken if the library is present on the host

```
$ ldd build/build_static/bin/aseprite
  linux-vdso.so.1 (0x00007fff72128000)
  libxcb.so.1 => /lib64/libxcb.so.1 (0x00007ff5658e2000)
  libGL.so.1 => /lib64/libGL.so.1 (0x00007ff56585b000)
  libfontconfig.so.1 => /lib64/libfontconfig.so.1 (0x00007ff56580c000)
  libSM.so.6 => /lib64/libSM.so.6 (0x00007ff565802000)
  libICE.so.6 => /lib64/libICE.so.6 (0x00007ff5657e4000)
  libX11.so.6 => /lib64/libX11.so.6 (0x00007ff56569d000)
  libXext.so.6 => /lib64/libXext.so.6 (0x00007ff565687000)
  libXcursor.so.1 => /lib64/libXcursor.so.1 (0x00007ff56567a000)
  libcrypto.so.3 => /lib64/libcrypto.so.3 (0x00007ff565200000)
  libacl.so.1 => /lib64/libacl.so.1 (0x00007ff565670000)
  libssl.so.3 => /lib64/libssl.so.3 (0x00007ff56515d000)
  libstdc++.so.6 => /lib64/libstdc++.so.6 (0x00007ff564e00000)
  libm.so.6 => /lib64/libm.so.6 (0x00007ff56507c000)
  libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007ff56564a000)
  libc.so.6 => /lib64/libc.so.6 (0x00007ff564a00000)
  /lib64/ld-linux-x86-64.so.2 (0x00007ff565935000)
  libXau.so.6 => /lib64/libXau.so.6 (0x00007ff565644000)
  libGLX.so.0 => /lib64/libGLX.so.0 (0x00007ff564dcd000)
  libGLdispatch.so.0 => /lib64/libGLdispatch.so.0 (0x00007ff564d15000)
  libfreetype.so.6 => /lib64/libfreetype.so.6 (0x00007ff564c45000)
  libxml2.so.2 => /lib64/libxml2.so.2 (0x00007ff56488c000)
  libuuid.so.1 => /lib64/libuuid.so.1 (0x00007ff565638000)
  libXrender.so.1 => /lib64/libXrender.so.1 (0x00007ff56562c000)
  libXfixes.so.3 => /lib64/libXfixes.so.3 (0x00007ff565074000)
  libz.so.1 => /lib64/libz.so.1 (0x00007ff56505a000)
  libattr.so.1 => /lib64/libattr.so.1 (0x00007ff564884000)
  libbz2.so.1 => /lib64/libbz2.so.1 (0x00007ff564870000)
  libpng16.so.16 => /lib64/libpng16.so.16 (0x00007ff564837000)
  libharfbuzz.so.0 => /lib64/libharfbuzz.so.0 (0x00007ff564739000)
  libbrotlidec.so.1 => /lib64/libbrotlidec.so.1 (0x00007ff56472c000)
  liblzma.so.5 => /lib64/liblzma.so.5 (0x00007ff5646f9000)
  libglib-2.0.so.0 => /lib64/libglib-2.0.so.0 (0x00007ff5645b0000)
  libgraphite2.so.3 => /lib64/libgraphite2.so.3 (0x00007ff56458f000)
  libbrotlicommon.so.1 => /lib64/libbrotlicommon.so.1 (0x00007ff56456c000)
  libpcre2-8.so.0 => /lib64/libpcre2-8.so.0 (0x00007ff5644d2000)
```

PR #64, opened 2023-08-04
Fix incorrect padding between frame images when packing sprite sheets (fix aseprite/aseprite#3993)

Gasparoken:
Prior to this fix, Export Sprite Sheet incorrectly generated a shape padding in some frame image when the following options were chosen:
- Packed sheet type
- Trim Cels
- Spacing > 0

fix aseprite/aseprite#3993


Gasparoken:
Pending: tests and border padding fix

martincapello (approved):

Gasparoken:
@dacap I think, ready to merge.

PR #66, opened 2023-08-16
Fix creation of cursor mask 

martincapello:
Fix aseprite/aseprite#3989

Before this change, on Windows, we were returning a "no cursor" from makeCursor when the cursor bitmap was fully transparent to avoid showing a black bitmap as the cursor. But it wasn't until recently that an issue was reported about the mouse cursor leaving a trail on the screen only when it was switched to a fully transparent bitpmap (see the fixed issue for details).
By always setting a cursor mask full of ones we avoid both of the aforementioned issues.

### DIFF
--- a/os/win/system.cpp
+++ b/os/win/system.cpp
@@ -117,8 +117,20 @@ public:
 
   // Create an empty mask bitmap.
   HBITMAP hmonobmp() {
-    if (!m_hmonobmp)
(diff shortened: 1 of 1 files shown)


dacap:
I was able to reproduce the bug and give a try to the fix 👍 it's working correctly.

PR #69, opened 2023-09-28
Fix: Drawing using a keyboard key on X11

tetektoza:
As spoken in https://github.com/aseprite/aseprite/pull/4063 - currently, if we try to bind a key to mouse trigger button and we will try to draw with it - after a longer period of holding the key, drawing will stop and will try to draw continously with short breaks.

This is caused because we never set repeat count inside the event that we generate from the event that comes from X Window System.

Repeat count is being properly checked in doc_view.cpp:onProcessMessage method, although it always equals 0 - thus allowing for the key event to be converted to mouse event, and at the same time interpreting (and queueing) it as a mouse event.

Solution for that is to simply check if the previous key event that came was KeyPress, and if it is currently pressed. If yes - it means that user never pushed the key down, thus set repeat count to 1. Otherwise, if the previous key event is KeyRelease for example - it means user has pushed the key down, thus repeat count on the current event can be set to 0.

### DIFF
--- a/os/x11/window.cpp
+++ b/os/x11/window.cpp
@@ -1042,7 +1042,19 @@ void WindowX11::processX11Event(XEvent& event)
         KEY_TRACE("Xutf8LookupString %s\n", &buf[0]);
       }
 
-      // Key event used by the input method (e.g. when the user
+      // Check if the key has been pressed, 
+      // and if yes - check if it's the same one key
+      // as in the previous event.  If it's the same one
+      // - it means key is being held, so set repeat to 1.
+      if (event.type == KeyPress) {
+        if (keysym == m_pressedKeySym)
+          ev.setRepeat(1);
+        m_pressedKeySym = keysym;
+      }
+      else
+        m_pressedKeySym = 0;
+
+      // Key event used by the input method (e.g. when the users
       // presses a dead key).
       if (XFilterEvent(&event, m_window))
         break;
--- a/os/x11/window.h
+++ b/os/x11/window.h
@@ -128,6 +128,8 @@ private:
   bool m_resizable = false;
   bool m_transparent = false;
 
+  KeySym m_pressedKeySym = 0;
+
   // Double-click info
   Event::MouseButton m_doubleClickButton;
   base::tick_t m_doubleClickTick;


tetektoza:
Maybe we could also change repeat variable to be of type bool, but I'd have to check in places where it's used so it would not cause a regression.

os/x11/window.h
```diff
@@ -108,6 +108,8 @@ class WindowX11 : public Window {
   bool setX11Cursor(::Cursor xcursor);
   bool requestX11FrameExtents();
   void getX11FrameExtents();
+  XEvent checkPrevKeyEvent() { return m_prev_key_event; }
+  void setPrevKeyEvent(XEvent curr_key_event) { m_prev_key_event = curr_key_event; }
```
martincapello, line 112: We can remove these private methods and access the member directly.

os/x11/window.cpp
```diff
@@ -1042,6 +1042,20 @@ void WindowX11::processX11Event(XEvent& event)
...
+      // has the same symbolic name as the previous key. If yes,
+      // then it means this button is being held down
+      XEvent prev_event = checkPrevKeyEvent();
+      if (prev_event.type != 0 && event.type == KeyPress) {
+        KeySym prev_keysym = XLookupKeysym(&prev_event.xkey, 0);
+        if (prev_event.type == KeyPress && x11_is_key_pressed(x11_keysym_to_scancode(prev_keysym)) && prev_keysym == keysym)
+          ev.setRepeat(1);
+      }
+
+      setPrevKeyEvent(event);
```
martincapello, line 1057: I think we could store the previous pressed KeySym instead of the XEvent and then refactor the code a bit:
 
```suggestion
      if (event.type == KeyPress) {
        if (m_pressedKeySym == keysym)
          ev.setRepeat(1);
        m_pressedKeySym = keysym;
      }
      else 
        m_pressedKeySym = 0;
```

os/x11/window.h
```diff
@@ -128,6 +130,8 @@ class WindowX11 : public Window {
   bool m_resizable = false;
   bool m_transparent = false;
 
+  XEvent m_prev_key_event = { 0 };
```
martincapello, line 133: As suggested in another comment, I think we should change this to KeySym m_pressedKeySym = 0;

martincapello (commented):
The fix works as expected, I'm just suggesting some changes because I think we can simplify a bit the code. Anyway, David will take a look and he might introduce more changes as well.

os/x11/window.h
tetektoza: Done as suggested.

os/x11/window.cpp
tetektoza: Done as suggested, right - it will simplify logic a lot.

os/x11/window.h
tetektoza: Done.

martincapello (commented):
LGTM!

dacap:
Thanks @tetektoza, I'll merge this as it works better than the current version. Anyway there are still problems when we use Shift modifier to keep a square ratio, etc on line/circle/rectangle tools.

As a side note, the idea of the "repeat" field was as a counter of "how many times the key was autopressed by the OS", but I'll make a comment about this in #70 where it makes more sense.

PR #70, opened 2023-10-01
Fix: Drawing using keys on Windows

tetektoza:
This patch is essentially the same implementation as in https://github.com/aseprite/laf/pull/69 - now the repeat count is being set correctly if we press the same event twice in a row, resulting in key events binded to mouse events being converted properly.

### DIFF
--- a/os/win/window.h
+++ b/os/win/window.h
@@ -138,6 +138,7 @@ namespace os {
     gfx::Rect m_restoredFrame;
 #endif
 
+    KeyScancode m_pressedKeyScancode = KeyScancode::kKeyNil;
     int m_scale;
     bool m_isCreated;
     // Since Windows Vista, it looks like Microsoft decided to change
--- a/os/win/window.cpp
+++ b/os/win/window.cpp
@@ -1679,7 +1679,13 @@ LRESULT WindowWin::wndProc(UINT msg, WPARAM wparam, LPARAM lparam)
       ev.setModifiers(get_modifiers_from_last_win32_message());
       ev.setScancode(ourScancode);
       ev.setUnicodeChar(0);
-      ev.setRepeat(std::max(0, int((lparam & 0xffff)-1)));
+
+      // if the last registered key scancode is equal to current
+      // scancode, it means the same event came, thus the key
+      // is being held down, so set repeat to 1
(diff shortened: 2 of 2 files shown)


os/win/window.h
```diff
@@ -138,6 +138,7 @@ namespace os {
     gfx::Rect m_restoredFrame;
 #endif
 
+    KeyScancode m_prev_keyscancode = KeyScancode::kKeyNil;
```
martincapello, line 141: Could you please change the name of this member to m_pressedKeyscancode?

martincapello (commented):
LGTM! Just suggested to change the member name, but the logic remains the same.

os/win/window.h
tetektoza: Done :)

dacap:
About this change there are some notes and an alternative fix that I'll like to push.

The idea of the `repeat` field was to match the repeat count of the `WM_KEYDOWN` event, but never worked as expected. From the "MSDN" documentation https://learn.microsoft.com/en-us/windows/win32/inputdev/wm-keydown about the LPARAM argument:

> Bits 0-15 = The repeat count for the current message. The value is the number of times the keystroke is autorepeated as a result of the user holding down the key.

And this is set here:

https://github.com/aseprite/laf/blob/592d028e5d9747c84dd846a166efc10e7363e98a/os/win/window.cpp#L1682

But the only problem is that it was not what I understood at the time. From [this answer](https://stackoverflow.com/a/44905726) the repeat count is the number of queued WM_KEYDOWN messages that are not processed, but if we process the queue fast enough, the value is always 0. It's not what I was expecting for a "autorepeat" count.

Anyway, we can still use one specific bit of the LPARAM which is the bit 30:

> Bit 30 = The previous key state. The value is 1 if the key is down before the message is sent, or it is zero if the key is up.

Which is already being used to detect if a key modifier is pressed for first time or it's being held:

https://github.com/aseprite/laf/blob/592d028e5d9747c84dd846a166efc10e7363e98a/os/win/window.cpp#L1670-L1675

So the alternative fix uses just this bit to set the repeat field: 58709a07c97243447897147c118c4e0005bbd162

dacap:
Thanks anyway @tetektoza about this, I'll update the laf submodule on aseprite repository now.

PR #71, opened 2023-10-03
Fix: Previous key state for repeat count in X11

tetektoza:
Currently, as we've spoken in https://github.com/aseprite/laf/pull/69 @dcap has correctly pointed out that my patch has some problems with shift key modifier - and he was correct, I've skipped an important thing looking into this problem.

So if we hold a key and the next event that will come will be different than the current one - it will reset repeat count. If it's shift for example, holding a trigger left mouse button on keyboard will correctly set the repeat count to 1, but as soon as we press shift during that for example
- it will bring the repeat count for the key to 0, since the new m_pressedKeySym will be equal to the shift button.

Solution for that is to simply keep a map of states of the current keyboard - it will remember last state of the button that got pressed. So even if we will press Shift, and after that comes another event with the key that got mapped to left mouse button - we will check the last state of the mapped key, not shift for example.


tetektoza:
The whole problem with the previous solution was that, that it naively interpreted previous pressed key symbol. So if we pressed another key while holding the mapped key, it reseted the repeat count in the mapped key to 0, since previous key symbol was reseted by the another key.

This still doesn't fully fixes the problem while holding Shift + V (if V is the key that got mapped to LMB), because right now the current implementation of Accelerator class (this is in Aseprite repo though) doesn't recognize mouse event if it comes with a modifier, i.e:

if we press Shift + V and rescale the square with aspect ratio for example, then if we stop holding V - it should stop resizing the square and just put it in place. This is what happens if you press LMB + Shift normally. It doesn't happen with keyboard shortcuts because accelerator can't interpret V as the KeyPress event, because it comes with a shift modifier.

I've prepared a solution for this in aseprite repo, will link it to this PR after I make a PR on aseprite.

os/x11/window.h
```diff
@@ -128,8 +128,7 @@ class WindowX11 : public Window {
   bool m_resizable = false;
   bool m_transparent = false;
 
-  KeySym m_pressedKeySym = 0;
-
+  std::map<KeySym, bool> map_keys;
```
dacap, line 131: I think it would be better if this is a globa/static/defined in the anonymous namespace of window.cpp, and it's called `g_pressedKeys`. And it might be an `std::set`.

os/x11/window.cpp
```diff
@@ -1042,18 +1042,18 @@ void WindowX11::processX11Event(XEvent& event)
...
       if (event.type == KeyPress) {
-        if (keysym == m_pressedKeySym)
+        if (map_keys[keysym])
           ev.setRepeat(1);
-        m_pressedKeySym = keysym;
       }
-      else
-        m_pressedKeySym = 0;
 
+      map_keys[keysym] = event.type == KeyPress ? true : false;
```
dacap, line 1056: I'm not sure if the map should be set to false when the key is released or just remove the element from the map.

I think we can have something like:
```c++
      if (event.type == KeyPress) {
        if (g_pressedKeys.find(keysym) != g_pressedKeys.end())
          ev.setRepeat(1);
        else
          g_pressedKeys.insert(keysym);
      }
      else
        g_pressedKeys.erase(keysym);
```

dacap (changes requested):

os/x11/window.h
tetektoza: Done, also please check out https://github.com/aseprite/aseprite/pull/4080 - this fixes a behaviour when we for example press Shift + Key that got mapped to mouse trigger button, and then we release the key that got mapped but never release Shift. I've explained it in the PR's comments.

os/x11/window.cpp
tetektoza: Done, also please check out https://github.com/aseprite/aseprite/pull/4080 - this fixes a behaviour when we for example press Shift + Key that got mapped to mouse trigger button, and then we release the key that got mapped but never release Shift. I've explained it in the PR's comments.

os/x11/window.h
```diff
@@ -24,7 +24,7 @@
 #include <X11/Xatom.h>
 #include <X11/Xlib.h>
 #include <X11/Xutil.h>
-
+#include <set>
```
dacap, line 27: This `#include` can go to the `window.cpp` file (because we aren't using `std::set` here in the .h)

tetektoza: Done, I've just seen it after force pushing first commit :D

os/x11/window.h
```diff
@@ -24,7 +24,6 @@
 #include <X11/Xatom.h>
 #include <X11/Xlib.h>
 #include <X11/Xutil.h>
-
```
dacap, line 27: Please don't remove this line.

tetektoza: Done, my bad. :)

dacap:
Thanks @tetektoza, I'll merge this patch and then give a try to https://github.com/aseprite/aseprite/pull/4080

PR #72, opened 2023-10-04
Flush pending events before opening pipe to zenity

Joefish:
Mouse clicks are ignored when we open a pipe to zenity without flushing the pending X11 events first because popen stalls aseprite execution until we selected a file.
This is in response to https://github.com/aseprite/aseprite/issues/3974

### DIFF
--- a/os/x11/native_dialogs.cpp
+++ b/os/x11/native_dialogs.cpp
@@ -16,6 +16,7 @@
 #include "base/replace_string.h"
 #include "base/split_string.h"
 #include "os/common/file_dialog.h"
+#include "os/x11/x11.h"
 
 #include <cstdio>              // popen/pclose()
 #include <cstring>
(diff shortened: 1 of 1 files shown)


os/x11/native_dialogs.cpp
```diff
@@ -139,8 +140,8 @@ class FileDialogX11 : public CommonFileDialog {
         // between the given "parent" argument in this show() function
         // and the created window by the utility.
 
-        // Here we run the command and get a handle to read its
-        // stdout.
+        XFlush(X11::instance()->display());
+        // Here we run the command and get a handle to read its stdout.
```
dacap, line 144: Thanks for your patch @Joefish, I'd recommend you to avoid touching the comment that is unrelated to this change and add a comment why XFlush() is needed:
```suggestion
        // ...Probably some comment why this is needed?...
        XFlush(X11::instance()->display());

        // Here we run the command and get a handle to read its
        // stdout.
```

Joefish: I restored the formatting of the comment and added a description for the XFlush line.

To be honest, when I tried opening a file with the native file dialog in aseprite, the mouse input was ignored/blocked and the cursor was still the same as when hovering over the 'open file' label in aseprite. So I concluded there must be pending events that haven't been sent to the X server yet before popen blocked further execution. And it worked, but is still just my best guess that this is the specific reason.

dacap:
Thanks a lot of this fix @Joefish, I'll merge it 👍

PR #74, opened 2023-11-30
Refactor external project arguments to use a list instead

martincapello:


martincapello:
You are right, it is not standard. I've named this way because they are arguments for the cmake command executed by the ExternalProject_Add function. I'll change it.

PR #75, opened 2023-12-26
Fix 'is_process_running' returns an incorrect 'true' when process 'pid' exists, but whose process does not belong to the current application.

Gasparoken:
Windows part not solved yet.
Linux not tested yet.

The current fix on all platforms might make aseprite/aseprite#4130 unnecessary.

### DIFF
--- a/base/process.h
+++ b/base/process.h
@@ -1,4 +1,5 @@
 // LAF Base Library
+// Copyright (c) 2023-2024 Igara Studio S.A.
 // Copyright (c) 2015-2016 David Capello
 //
 // This file is released under the terms of the MIT license.
@@ -15,6 +16,8 @@ namespace base {
   typedef uint32_t pid;
 
   pid get_current_process_id();
+
+  bool is_process_running(pid pid, const char* pname);
 
   bool is_process_running(pid pid);
 
--- a/base/process.h
+++ b/base/process.h
@@ -11,14 +11,20 @@
 
 #include "base/ints.h"
 
+#include <string>
+
 namespace base {
 
   typedef uint32_t pid;
 
   pid get_current_process_id();
 
-  bool is_process_running(pid pid, const char* pname);
+  std::string get_process_name(pid pid);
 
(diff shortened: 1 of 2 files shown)


Gasparoken:
Ready for review. Linux version without modifications.
This fix is needed to fix https://github.com/aseprite/aseprite/issues/4130.

martincapello (approved):
LGTM! I have added the Linux part.

Gasparoken:
I changed the focus. @martincapello could you please review it again and give me your opinion on the implementation? Could you try again on Linux?

base/process.h
```diff
@@ -16,7 +17,9 @@ namespace base {
 
   pid get_current_process_id();
 
-  bool is_process_running(pid pid);
+  std::string get_process_name(pid pid);
```
martincapello, line 20: I think you should include <string> to avoid the errors reported by Github actions.

martincapello (commented):
Looks better than before! I tested this on Linux and it works. However I'm not sure why when compiling it doesn't complain as the github actions do...I left a comment with a possible fix that should make github actions happy.

martincapello (approved):

Gasparoken:
@dacap the PR is ready for your review.

dacap (commented):
I've made some comments, the commits are a little messy (including git descriptions), I can take this PR from here refactoring/rebasing/editing all the commits in some way (I have 3 platforms running right now to implement this ASAP).

dacap:
Merged in:

* https://github.com/aseprite/laf/commit/a6c13ab3c349cf76232c7d048afb051cf2a82537 by @Gasparoken 
* https://github.com/aseprite/laf/commit/4e5249a84900607520de71b6097bde80bbe45469 by @martincapello 

I'll see if I change the Windows impl later but the returned value should be the same. With this new get_process_name() function I think the [original Aseprite bug](https://github.com/aseprite/aseprite/issues/4130#issuecomment-1866333196) could be fixed.

PR #76, opened 2024-02-08
Fix first glyph clipping when drawing text with FreeType fonts

martincapello:
Fix aseprite/aseprite#4199


dacap:
Just tested, it works great 👍

PR #82, opened 2024-02-28
Add mouse coordinates to os::Event::DropFiles event indicating the point where the drop of files took place on the target window

martincapello:
Implements pending work needed for https://github.com/aseprite/aseprite/issues/131

### DIFF
--- a/examples/allevents.cpp
+++ b/examples/allevents.cpp
@@ -40,7 +40,9 @@ public:
         break;
 
       case os::Event::DropFiles:
-        logLine("DropFiles files={");
+        logLine("DropFiles pos=%d,%d files={",
+                ev.position().x,
+                ev.position().y);
         for (const auto& file : ev.files()) {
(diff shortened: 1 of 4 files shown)


dacap:
I've pushed other commit in the branch for the allevents example. I think we'll need a new kind of event, because DropFiles occurs when we release the mouse button, but for this we'll need some kind of feedback in the whole drag operation.

martincapello:
Great. Yeah, we will need a new event for that. DragFiles?

dacap:
That seems the best decision. I can merge this right now (as this PR adds the position of the DropFiles event correctly), and do the whole DnD work in a specific dnd branch for that. I could take some part of the implementation if you wish.

martincapello:
Makes sense. Merge this as is and I will continue working in a different branch. First I would like to start drafting some implementation for macOS to agree with you about the api that should be exposed to laf clients and some implementation details.

Issue #11, opened 2019-07-20
Prepare API for the first release


dacap:
Some changes that must be done for a probable first release of the library:

- [ ] os::Surface::bounds() -> Rect(0, 0, width, height)
- [ ] os::Display::invalidate() = os::Display->invalidateRegion(display->surface->bounds)
- [ ] rename setTitleBar() -> setTitle()
- [ ] getEvent() -> replace bool arg with an enum (WaitForEvents, DontWait)
- [ ] what to do with os::ScopedHandle? add os::DisplayHandle and os::SurfaceHandle aliases?
- [ ] ...

PR #83, opened 2024-03-04
Introduce drag and drop handling

martincapello:
This PR contains the drag and drop implementation in each platform for laf. 
Also exposes the API that can be used by laf clients to use drag and drop in a platform-independent way.


dacap:
Although it's not a bad idea to use these kind of delegates, I think it will make a little harder to process the events in the same order in Aseprite `ui::Manager::generateMessagesFromOSEvents()`.

I think it would be better to keep all event handling as a queue of events (so we know the exact order of events).

dacap:
It looks like a similar case to `Window::handleHitTest` or `System::handleWindowResize` (you will see a comment about a possible future `SystemDelegate` to replace `handleWindowResize`). These public `std::function` fields should be replaced in a future with some kind of delegate (probably everything in the window as a possible `WindowDelegate`).

We could offer some others `handleDndSomething` function fields but doing the delegate solution directly will be the best for the whole drag and drop mechanism. We can give a try and if it does work leave it as it is. (And if it doesn't work, then refactor it with events.)

I was also thinking that the drag and drop events will require several new fields that regular `os::Event`s don't (and the queue of events is an array of `os::Event`s), so probably it will make sense to just create a `os::DndEvent` for all these member functions of the delegate with all the drag and drop info required.

Some comments:
1. Don't base your example on `multiple_windows`, `drag_and_drop` example could just have one window (make the example as small as possible to show only the drag and drop mechanism)
2. Not sure if using `DnD` or `Dnd` as prefix or things like `DragSource` and `DropTarget` as class names for the delegates (same for the event, it could be `os::DragEvent`)
3. The header file could be called `os/dnd.h` including everything needed for drag and drop
4. `os/os.h` should include `os/dnd.h` so examples don't require to include this specific header 
5. I'm doing several refactors on the beta branch, this shouldn't be a problem as I'll merge/fix conflicts later with your changes

martincapello:
> It looks like a similar case to `Window::handleHitTest` or `System::handleWindowResize` (you will see a comment about a possible future `SystemDelegate` to replace `handleWindowResize`). These public `std::function` fields should be replaced in a future with some kind of delegate (probably everything in the window as a possible `WindowDelegate`).

Yeah, I saw those and I imagined that there were there because of a similar reason, in fact I've considered using functions instead of the delegates, but I think I saw the comment you mention so I decided to go for delegates at the end.

> We could offer some others `handleDndSomething` function fields but doing the delegate solution directly will be the best for the whole drag and drop mechanism. We can give a try and if it does work leave it as it is. (And if it doesn't work, then refactor it with events.)

Ok, I'll go ahead with the proposed delegates then.

> I was also thinking that the drag and drop events will require several new fields that regular `os::Event`s don't (and the queue of events is an array of `os::Event`s), so probably it will make sense to just create a `os::DndEvent` for all these member functions of the delegate with all the drag and drop info required.

Alright, will create a new os::DndEvent class that doesn't inherit from os::Event.

> Some comments:
> 
> 1. Don't base your example on `multiple_windows`, `drag_and_drop` example could just have one window (make the example as small as possible to show only the drag and drop mechanism)

Sure, that is the idea. Making a copy of multiple_windows just was the quicker thing to do for me at the time.
 
> 2. Not sure if using `DnD` or `Dnd` as prefix or things like `DragSource` and `DropTarget` as class names for the delegates (same for the event, it could be `os::DragEvent`)

Sure. We could also create a dnd namespace I guess, and use Source and Target as class names. But it is up to you, I don't have any preference. Just let me know if you want to go ahead with the names you proposed.

> 3. The header file could be called `os/dnd.h` including everything needed for drag and drop

Sounds good.

> 4. `os/os.h` should include `os/dnd.h` so examples don't require to include this specific header

Perfect.

> 5. I'm doing several refactors on the beta branch, this shouldn't be a problem as I'll merge/fix conflicts later with your changes

👍

martincapello:
@dacap This is still in draft but I want you to take a look at the current status. You can run the drag_and_drop example (which is not finished yet) to see some of things in action. 
Things implemented (just for macOS right now):
- DragTarget definition.
- Realtime drag feedback when the user moves the dragged data.
- Support for dropping file paths from finder (this is similar as we have in the main branch, but using the DragTarget interface instead).
- Introduction of DragDataProvider interface and its implementation for macOS. This is used to get data from the dragged payload, i.e. the Pasteboard in macOS.

Things missing:
- DragSource definition.
- Support for other kind of data items (images, text, pdf, RTF, strings, etc).
- Changing the cursor at will.

However, I think that with the current status (and once we got this implemented in Windows and Linux) we have the basic blocks in place to start with https://github.com/aseprite/aseprite/issues/131 in Aseprite. Then we can implement the rest in laf in the future. What do you think?

dacap:
The first thing I tried was to drag this box and didn't work:

Screenshot 2024-03-13 at 14 26 15

I'll try to review this soon.

martincapello:
Hehe...yeah, it can't be dragged...the DragSource is not implemented. So just try to drop files from finder...or any other DnD source app that provides NSFilenamesPboardType items.

dacap:
Probably we should create an example showing only the functionality we're implementing/need for the feature. Drop paths and images, nothing else.

martincapello:
Yes, I agree. Once we are okay with the approach taken I will remove the things that are not functional from the example. Images are not supported at the moment. There are some commented code just to show how they could be approached, but nothing else.

martincapello:
My idea was to implement the minimum necessary for supporting drag & dropping of files with realtime update of the UI while dragging them.

os/dnd.h
```diff
@@ -0,0 +1,125 @@
...
+      if constexpr (IT == Paths) {
+        return m_dataProvider->getPaths();
+      }
+      /*
+      else if constexpr (IT == Images) {
+        return m_dataProvider->getImages();
+      }
+      */
+      return T();
+    }
```
dacap, line 84: I'd remove this and just use the DragDataProvider directly in the client side (we need a getter for the m_dataProvider). And the DragDataProvider a way to check the type of data/or included types in the data (?)

dacap (changes requested):
So far it looks good, some minor changes to the API and still some decisions to take (e.g. should we use unique pointers for targets/sources or leave the owning decision to the client side).

The example could be simplified just showing a box where we can drop things (files/images) and accept or not the DragEvent if we drop things inside that box (to avoid switching flags so many flags).

dacap:
About data types, we need only filenames/paths and images. Generally the "drag and drop" functions are related to the clipboard functions, so it might be possible that some functions from the clip library can be used (which might require changes in the clip library).

dacap:
One last time at the moment is if we can merge dropResult/acceptDrop in some way, or explain when each field must be set/used on each event (and if we need a DropEvent class to separate fields for a drop-only event).

os/window.h
```diff
@@ -204,11 +211,18 @@ namespace os {
     virtual GrDirectContext* sk_grCtx() const = 0;
 #endif
 
+    virtual void onDragEnter(os::DragEvent& ev);
+    virtual void onDrag(os::DragEvent& ev);
+    virtual void onDragLeave(os::DragEvent& ev);
+    virtual void onDrop(os::DragEvent& ev);
+
```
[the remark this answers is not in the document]
martincapello: Sure, it makes sense. I'll make the changes.

os/window.h
```diff
@@ -194,6 +197,10 @@ namespace os {
     // devices like stylus can be used to move and resize the window.
     std::function<Hit(os::Window*, const gfx::Point& pos)> handleHitTest = nullptr;
 
+    void setDragSource(std::unique_ptr<DragSource> delegate) { m_dragSource = std::move(delegate); }
+    void setDragTarget(std::unique_ptr<DragTarget> delegate) { m_dragTarget = std::move(delegate); }
```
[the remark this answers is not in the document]
martincapello: I wasn't sure either...but now I think that a raw pointer give us more flexibility. So I'll change it.

martincapello:
> One last time at the moment is if we can merge dropResult/acceptDrop in some way, or explain when each field must be set/used on each event (and if we need a DropEvent class to separate fields for a drop-only event).

We can remove the acceptDrop field and let DragTarget::drop return a boolean that indicates if the drop was accepted or not. Something similar could be done for DragTarget::dragEnter and DragTarget::draggingUpdated by making them return a DropOperation and then removing the dropResult field from DragEvent.

dacap:
> We can remove the acceptDrop field and let DragTarget::drop return a boolean that indicates if the drop was accepted or not. Something similar could be done for DragTarget::dragEnter and DragTarget::draggingUpdated by making them return a DropOperation and then removing the dropResult field from DragEvent.

martincapello:
> About data types, we need only filenames/paths and images. Generally the "drag and drop" functions are related to the clipboard functions, so it might be possible that some functions from the clip library can be used (which might require changes in the clip library).

I've been looking the clip library, and yes, all the image-related functions could be adapted to reuse them in laf. Will see what can I do.

os/skia/skia_surface.cpp
```diff
@@ -326,6 +373,89 @@
...
+    default:
+      throw base::Exception("Unsupported alpha type");
+  }
+}
+
+// static
+SkAlphaType SkiaSurface::asSkAlphaType(const os::PixelAlpha& pa)
+{
+  switch(pa) {
+    case os::PixelAlpha::kOpaque:
```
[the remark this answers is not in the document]
martincapello: 🤌

martincapello:
Made the suggested changes and added the ability to get images from the dragged data. For this I copied and adapted code from the clip library (for macOs) to use laf surfaces. It was also needed to add some things to laf.

os/dnd.h
```diff
@@ -0,0 +1,106 @@
...
+    Paths,
+    Image
+  };
+
+  // Interface to get dragged data from the platform's implementation.
+  class DragDataProvider {
+  public:
+    virtual ~DragDataProvider() {}
+    virtual base::paths getPaths() = 0;
+    virtual std::unique_ptr<os::Surface> getImage() = 0;
```
dacap, line 46: We can use the `SurfaceRef` directly here because `Surface` is a `RefCount` object (has a reference counter included):
```suggestion
    virtual os::SurfaceRef getImage() = 0;
```

os/surface_format.h
```diff
@@ -8,12 +9,20 @@
...
 
+#include <cstdint>
+
 namespace os {
 
   enum SurfaceFormat {
     kRgbaSurfaceFormat,
   };
 
+  enum PixelAlpha {
```
dacap, line 20: Prefer enum class (we should do the same for SurfaceFormat in a future, it's an old enum):
```suggestion
  enum class PixelAlpha {
```

os/skia/skia_surface.h
```diff
@@ -115,6 +118,9 @@ class SkiaSurface final : public Surface {
   static SurfaceRef loadSurface(const char* filename);
 
 private:
+  static SkColorType deductSkColorType(const os::SurfaceFormatData& sf);
+  static os::PixelAlpha asPixelAlpha(const SkAlphaType& at);
+  static SkAlphaType asSkAlphaType(const os::PixelAlpha& pa);
```
dacap, line 123: All functions that receive an enum, can receive it as a value instead of `const T&`, because it's an integer.
```suggestion
  static os::PixelAlpha asPixelAlpha(SkAlphaType at);
  static SkAlphaType asSkAlphaType(os::PixelAlpha pa);
```
e.g. Without this change you cannot use asPîxelAlpha() with a literal value.

os/surface_format.h
martincapello: I made a separate commit making SurfaceFormat an enum class.

martincapello:
Just pushed some progress on the windows implementation. It is not finished and also broke something in the macOs implementation because of some changes that I didn't have the chance to make today.

os/win/dnd.h
```diff
@@ -0,0 +1,96 @@
...
+
+#include "os/dnd.h"
+
+#include <ole2.h>
+
+namespace os {
+
+  // COM object pointer wrapper to simplify reference counting.
+  template<class T>
+  class ComPtr {
```
dacap, line 19: We have a COM pointer wrapper in `base::ComPtr` (`#include "base/win/comptr.h"`)

dacap:
As we talked the other time, probably we can make the laf library depends on the clip library. I can prepare this change on aseprite and laf main branch moving the clip submodule from aseprite to laf.

martincapello:
> As we talked the other time, probably we can make the laf library depends on the clip library. I can prepare this change on aseprite and laf main branch moving the clip submodule from aseprite to laf.

Sure, but the thing was that the code in the clip library doesn't use os::Surface. So I think I will have to figure out how to interface from laf to the clip library without making clip aware of anything from laf.

dacap:
> the thing was that the code in the clip library doesn't use os::Surface. So I think I will have to figure out how to interface from laf to the clip library without making clip aware of anything from laf.

Yes, but that shouldn't be a problem, as the clip library has a 32-bit image format with custom mask/shift (just like the os::Surface, the same pixel data could be used directly between clip/laf-os).

I'll try to move the submodule between the projects so we have clip already available in laf.

dacap:
The `clip` module is available in the `main` branch now.

martincapello:
Oh, just in case, in the clip library there is another comptr implementation in clip_win_wic.h...which seems like incomplete or something (for instance it doesn't call AddRef() in any method).

dacap:
> Oh, just in case, in the clip library there is another comptr implementation in clip_win_wic.h...which seems like incomplete or something (for instance it doesn't call AddRef() in any method).

The clip library is an independent library and has its own wrappers (comptr, coinit, hmodule, Hglobal) with all the methods that it needs. AFAIR as the clip comptr is a wrapper only to release created resources, there is no need for add ref (the reference is added by each creation functions, e.g. CoCreateInstance() adds the reference, and we then release it).

In the best case, we should avoid any interaction between laf <-> clip using COM pointer wrappers (just use raw pointers between boundaries if it's necessary, just like any other COM API).

base/win/comptr.h
```diff
@@ -31,6 +33,14 @@ namespace base {
...
     }
 
+    ComPtr(T* p) {
+      if (!p)
+        return;
+
+      if (p->QueryInterface<T>(&m_ptr) != NOERROR)
+        m_ptr = nullptr;
+    }
+
```
dacap, line 43: As `T` is the same as the `T*p`, I think we can simplify this as
```suggestion
    explicit ComPtr(T* p) : m_ptr(p) {
      if (m_ptr)
        m_ptr->AddRef();
    }

```
I have added the `explicit` keyword to avoid implicit conversion between raw pointers and this class adding new references/querying interfaces.

martincapello: This simplification actually skips the [QueryInterface](https://learn.microsoft.com/en-us/windows/win32/api/unknwn/nf-unknwn-iunknown-queryinterface(q)) call...which I actually wanted to call because of the validations it does. Do you they are not needed?

dacap: If you are copying/adding a new reference to a COM object pointer of the same type (T*) the `QueryInterface()` is not needed (shouldn't), as the pointer to `T` interface should be already valid. (`QueryInterface` is used to ask for another interface.)

Unrelated to this: The only problem with this kind of constructor (`ComPtr(T*)`) is that we might want two kind of behaviors: 1) adding a new reference (with AddRef), or 2) wrapping an existent reference to do just the automatic release (e.g. as clip comptr).

Anyway the `AddRef()` behavior is the most common, and in case you need one without AddRef we could add something like an attach method (e.g. https://github.com/microsoft/wil/blob/5c6a7ba43e79edd71dcd1ce4c2d21b0610183c86/include/wil/com.h#L360 ).

dacap:
I fear that this PR might be impossible to review if gets to big. It might be interesting to think an approach were we merge first the macOS impl, then the Windows one, and finally the Linux one. (Even if some refactors to the events must be done for each impl.)

martincapello:
> I fear that this PR might be impossible to review if gets to big. It might be interesting to think an approach were we merge first the macOS impl, then the Windows one, and finally the Linux one. (Even if some refactors to the events must be done for each impl.)

Sure, no problem. But I think that first I would like to merge the windows implementation. Because I'm currently working on it along the some clip lib changes (since now laf depends on clip). Then I will have to refactor some things of the macOS impl to make it use clip (instead of the copy & paste I did before the clip dependency).

martincapello:
Okay, I've organized a bit the commits hopefully. These changes need this PR merged into aseprite fork: https://github.com/dacap/clip/pull/69 

Also, the macOS implementation was removed for now, I have to make some adjustments and then I will commit ir separately, maybe in a different PR after we've merged this.

martincapello:
Had to force push the last two commits to introduce the changes mentioned here: https://github.com/dacap/clip/pull/69#issuecomment-2050846596

martincapello:
Sorry to force push again, but it really makes sense to me to have that clip submodule update within the windows implementation changes.

dacap:
I'll start reviewing this right now. So in some way I don't expect new commits until it's reviewed or merged.

dacap:
@martincapello how can I test dropping an image on Windows? it looks like it's not working, I'm trying to drop an image from Chrome.

Also I'd change the location of the drop zone to this corner (on macOS tested dropping an image and it might move the drop zone out of the window area):

image

os/dnd.h
```diff
@@ -0,0 +1,108 @@
...
+#pragma push_macro("None")
+#undef None // Undefine the X11 None macro
+
+namespace os {
+
+  class Window;
+
+  // Operations that can be supported by source and target windows in a drag
+  // and drop operation.
+  enum DropOperation {
```
dacap, line 27: This should be an enum class:
```suggestion
  enum class DropOperation {
```

os/dnd.h
```diff
@@ -0,0 +1,108 @@
...
+    Paths,
+    Image,
+  };
+
+  // Interface to get dragged data from the platform's implementation.
+  class DragDataProvider {
+  public:
+    virtual ~DragDataProvider() {}
+    virtual base::paths getPaths() = 0;
+    virtual SurfaceRef getImage() = 0;
```
dacap, line 46: In some future commits, we could add support to get all the dropped types/raw data.

os/dnd.h
```diff
@@ -0,0 +1,108 @@
...
+    // do not combine values using bitwise operators.
+    void dropResult(DropOperation operation) { m_dropResult = operation; }
+    // Set this to true when the dropped data was accepted/processed by the
+    // target window, or set to false otherwise.
+    void acceptDrop(bool value) { m_acceptDrop = value; }
+
+    bool sourceSupports(DropOperation op) { return (m_supportedOperations & op) == op; }
+
+  private:
+    os::Window* m_target;
```
dacap, line 80: Just in case:
```suggestion
    os::Window* m_target = nullptr;
```

os/osx/dnd.h
```diff
@@ -0,0 +1,129 @@
...
+#include "os/dnd.h"
+#include "os/surface.h"
+#include "os/surface_format.h"
+#include "os/system.h"
+
+#include <Cocoa/Cocoa.h>
+#include <memory>
+
+namespace os {
+  class DragDataProviderOSX : public DragDataProvider {
```
dacap, line 28: Move the implementation of this class and functions inside this file in a new .mm file.

martincapello:
> @martincapello how can I test dropping an image on Windows? it looks like it's not working, I'm trying to drop an image from Chrome.
>

As I mentioned [here](https://github.com/aseprite/laf/pull/83#issuecomment-2049786443) on Windows you have to use some app that let you drag images, I had to create one for this because Chrome doesn't do the same as in macOS.

> Also I'd change the location of the drop zone to this corner (on macOS tested dropping an image and it might move the drop zone out of the window area):
> 
> image

Sure, I'll change it.

os/skia/skia_surface.cpp
```diff
@@ -76,6 +82,40 @@ void SkiaSurface::create(int width, int height, const os::ColorSpaceRef& cs)
...
+
+  if (data) {
+    // Convert 24bpp data to 32bpp data because Skia doesn't work with 24bpp
+    // images.
+    if (sf.bitsPerPixel == 24) {
+      for (int i=0,j=0; j<width * height * 3; i+=4,j+=3)
+        std::memcpy(&static_cast<unsigned char*>(bmp.getPixels())[i], &data[j], 3);
+    }
+    else
+      bmp.writePixels(SkPixmap(bmp.info(), data, size_t(width) * size_t(sf.bitsPerPixel / 8)));
```
dacap, line 111: I'm not sure if it's clear enough what the `sf` parameter means, is the format of the resulting SkiaSurface, or the format of the input `data` buffer? Probably we need an extra parameter for the input `data` format.

dacap:
> As I mentioned [here](https://github.com/aseprite/laf/pull/83#issuecomment-2049786443) on Windows you have to use some app that let you drag images, I had to create one for this because Chrome doesn't do the same as in macOS.

I think it should work with some standard program at least (Chrome, Edge, etc.).

os/skia/skia_surface.cpp
```diff
@@ -326,6 +367,89 @@ void SkiaSurface::getFormat(SurfaceFormatData* formatData) const
...
+      sf.greenMask  == SK_G16_MASK &&
+      sf.blueMask   == SK_B16_MASK &&
+      sf.alphaMask  == 0)
+    return kRGB_565_SkColorType;
+
+  return kUnknown_SkColorType;
+}
+
+// static
+os::PixelAlpha SkiaSurface::asPixelAlpha(SkAlphaType at)
```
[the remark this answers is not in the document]
dacap: Check this comment of clang-tidy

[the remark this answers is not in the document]
dacap: Just in case, it means that the static private function might be just static non-member functions (no need to put them in the header file.)

os/win/dnd.cpp
```diff
@@ -0,0 +1,335 @@
...
+
+  return static_cast<os::DropOperation>(op);
+}
+
+
+gfx::Point drag_position(HWND hwnd, POINTL& pt)
+{
+  ScreenToClient(hwnd, (LPPOINT) &pt);
+  return gfx::Point(pt.x, pt.y);
+}
```
dacap, line 51: All these functions need to be ´static´ or in a anonymous namespace.

os/dnd.h
martincapello: Well...actually I've tried to use the enum class for this...but since this are flags and it is designed to be used with bitwise operators...I found it pretty inconvenient to have to implement all the bitwise operators to make it work as I wanted. So I opted to leave it as an enum. If you prefer the enum class along with the bitwise operators implementation, let me know and I can do it. Or maybe there is some other way to do this that I'm not aware of.

dacap: Yes, `enum class` as flags are hard to use (a lot of `static_cast<int>` are required), but this must be done (in other case we have things like `os::Copy` or `os::None` available in the whole `os` namespace).

os/win/dnd.cpp
```diff
@@ -0,0 +1,335 @@
...
+  }
+
+  SIZE_T size() {
+    return GlobalSize(m_hmem);
+  }
+
+private:
+  HGLOBAL m_hmem;
+  T m_data;
+};
```
dacap, line 84: I think `T` should be the type without the pointer (just like other pointer-like wrappers/smart-pointers):
```suggestion
  GLock(HGLOBAL hglobal) : m_hmem(hglobal) {
    m_data = static_cast<T*>(GlobalLock(m_hmem));
  }

  ~GLock() {
    GlobalUnlock(m_hmem);
  }

  operator HGLOBAL() {
    return m_hmem;
  }

  operator T*() {
    return m_data;
  }

  SIZE_T size() {
    return GlobalSize(m_hmem);
  }

private:
  HGLOBAL m_hmem;
  T* m_data;
};
```

os/win/dnd.cpp
```diff
@@ -0,0 +1,335 @@
...
+    {
+      GLock<HDROP> hdrop(medium.hGlobal);
+      if (static_cast<HDROP>(hdrop)) {
+        int count = DragQueryFile(hdrop, 0xFFFFFFFF, nullptr, 0);
+        for (int index = 0; index < count; ++index) {
+          int length = DragQueryFile(hdrop, index, nullptr, 0);
+          if (length > 0) {
+            std::vector<TCHAR> str(length + 1);
+            DragQueryFile(hdrop, index, &str[0], str.size());
+            files.push_back(base::to_utf8(&str[0]));
```
dacap, line 106: We can add a comment about this `length+1`:
```suggestion
            // From Win32 docs: https://learn.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-dragqueryfilew
            // the DragQueryFile() does't include the null character in its return value.
            std::vector<TCHAR> str(length + 1);
            DragQueryFile(hdrop, index, str.data(), str.size());
            files.push_back(base::to_utf8(str.data()));
```
I have changed `&str[0]` to `str.data()` because clang-tidy prefer that (this file was't clang-tidy'd because it runs in the Linux port only).

martincapello:
> > As I mentioned [here](https://github.com/aseprite/laf/pull/83#issuecomment-2049786443) on Windows you have to use some app that let you drag images, I had to create one for this because Chrome doesn't do the same as in macOS.
> 
> I think it should work with some standard program at least (Chrome, Edge, etc.).

Then we should add support for more Clipboard formats on windows. Because if Chrome (or Edge) only sends the URL of the image, we will need to support getting the URL, and the work of getting the actual image data will be responsibility of the library client. Right?

os/win/dnd.cpp
```diff
@@ -0,0 +1,335 @@
...
+    GLock<BITMAPINFO*> hbi(medium.hGlobal);
+    clip::win::BitmapInfo bi(hbi);
+    if (bi.to_image(img)) {
+      spec = img.spec();
+      goto makeSurface;
+    }
+  }
+
+  // No suitable image format found, release medium and return.
+  goto releaseMedium;
```
dacap, line 164: This indicates that probably we need a wrapper (like `GLock`) to automatically call `ReleaseStgMedium` in its dtor and avoid two goto labels in the same function.

Probably we can reorganize the logic of this function, e.g. creating other private `DragDataProviderWin` function.

os/win/dnd.cpp
```diff
@@ -0,0 +1,292 @@
...
+                          bitmap.pixelsHigh,
+                          sf,
+                          bitmap.bitmapData
+                        );
+  return surface;
+  */
+  return nullptr;
+}
+
+bool DragDataProviderWin::contains(DragDataItemType type)
```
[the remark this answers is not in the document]
dacap: A clang-tidy error. Not sure if it does check the Windows code anyway from Linux but fails as there are a lot of types/header files that are not found.

os/win/dnd.cpp
```diff
@@ -0,0 +1,335 @@
...
+        break;
+      case CF_DIB:
+        if (type == DragDataItemType::Image)
+          return true;
+        break;
+      default:
+        int namelen = GetClipboardFormatNameA(fmt.cfFormat, name, 100);
+        name[namelen] = '\0';
+        if (std::strcmp(name, "PNG") == 0 && type == DragDataItemType::Image)
+          return true;
```
dacap, line 210: Add a scope here:
```suggestion
      default: {
        int namelen = GetClipboardFormatNameA(fmt.cfFormat, name, 100);
        name[namelen] = '\0';
        if (std::strcmp(name, "PNG") == 0 && type == DragDataItemType::Image)
          return true;
        break;
      }
```

os/win/dnd.cpp
```diff
@@ -0,0 +1,335 @@
...
+                                          DWORD* pdwEffect)
+{
+  if (!m_window->hasDragTarget())
+    return E_NOTIMPL;
+
+  m_data = base::ComPtr<IDataObject>(pDataObj);
+  if (!m_data)
+    return E_UNEXPECTED;
+
+  m_position = drag_position((HWND) m_window->nativeHandle(), pt);
```
dacap, line 257:
```suggestion
  m_position = drag_position((HWND)m_window->nativeHandle(), pt);
```

dacap:
> Then we should add support for more Clipboard formats ~on windows~. Because if Chrome (or Edge) only sends the URL of the image, we will need to support getting the URL, and the work of getting the actual image data will be responsibility of the library client. Right?

If that is what Chrome reports in the drop operation we should do that (or report the drop operation as an URL data type, so the app can download the URL). I think this drag-and-drop from Chrome/web browser to the program will be the most common (and probably unique) use case dragging images directly into Aseprite.

os/win/system.cpp
```diff
@@ -155,9 +159,24 @@ SystemWin::SystemWin()
...
 }
 
+void SystemWin::setAppMode(AppMode appMode)
+{
+  m_appMode = appMode;
+  if (m_appMode == AppMode::GUI) {
+    auto result = OleInitialize(nullptr);
+    if (result != S_OK && result != S_FALSE) {
+      LOG(LogLevel::ERROR, "WIN: Could not initialize OLE (%d)", result);
+      return;
```
dacap, line 175: No need for this return as we'll return anyway and we should keep running if the `OleInitialize` fails.
```suggestion
```

os/dnd.h
martincapello: Okay, so should we use the static_casts? Or should I implement the bitwise operators methods?

os/win/system.cpp
```diff
@@ -155,9 +159,24 @@ SystemWin::SystemWin()
...
+    OleUninitialize();
+  }
   destroyInstance();
 }
 
+void SystemWin::setAppMode(AppMode appMode)
+{
+  m_appMode = appMode;
+  if (m_appMode == AppMode::GUI) {
+    auto result = OleInitialize(nullptr);
```
dacap, line 172:
```suggestion
    HRESULT result = OleInitialize(nullptr);
```

os/win/window.cpp
```diff
@@ -277,6 +277,11 @@ WindowWin::WindowWin(const WindowSpec& spec)
   // WindowSpec::activate() flag for this)
   m_activate = (spec.parent() == nullptr);
 
+  auto result = RegisterDragDrop(m_hwnd, reinterpret_cast<LPDROPTARGET>(&m_dragTargetAdapter));
```
dacap, line 280:
```suggestion
  HRESULT result = RegisterDragDrop(m_hwnd, reinterpret_cast<LPDROPTARGET>(&m_dragTargetAdapter));
```

os/win/dnd.cpp
```diff
@@ -0,0 +1,335 @@
...
+{
+  ScreenToClient(hwnd, (LPPOINT) &pt);
+  return gfx::Point(pt.x, pt.y);
+}
+
+namespace os {
+
+// HGLOBAL Locking/Unlocking wrapper
+template <typename T>
+class GLock {
```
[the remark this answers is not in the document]
martincapello: Which option do you prefer?

os/win/window.h
```diff
@@ -138,6 +139,8 @@ namespace os {
     gfx::Rect m_restoredFrame;
 #endif
 
+    DragTargetAdapter m_dragTargetAdapter = DragTargetAdapter(this);
```
dacap, line 142: One thing I don't like about this (and calling RegisterDragDrop) for every window is that there are windows (like menu popups/tooltips) that don't require this drag-and-drop logic.

dacap: Could we convert this to a pointer and initialize the drag-and-drop logic if the m_dragTarget is set?

os/win/dnd.cpp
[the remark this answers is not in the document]
dacap: Both are fine (you can choose the easier for you). If you move `GLock` to `base::win`, it will need another name.

os/dnd.h
dacap: Use static_cast when is needed.

os/skia/skia_surface.cpp
martincapello: It is actually for both. The data must be arranged/formatted according to the SurfaceFormatData passed. Then if Surface::getFormat(SurfaceFormatData* formatData) is called on the created surface, it should fill the formatData with the original values (except the `format` member which is not used to create the surface, but anyway there is only one value...maybe this member is the confusing part?).
Hence, I don't think that adding an extra parameter will make things clear. Maybe just documenting the method would be enough.

dacap: This is the confusing part: `sf.bitsPerPixel == 24`
This is the input data format, but not the output (the output will be 32 bpp, although not sure if kRGBA_8888_SkColorType or kBGRA_8888_SkColorType). I suppose here you are expecting the user to use the same shift/mask of RGB channels in 24bpp and 32bpp (?)

Anyway I think that if we expose the whole range of Skia formats in our laf-os API, we could completely remove the SurfaceFormatData. And in that case I'm not sure how the 24bpp would be specified for the input data.

dacap: Other possibility to simplify the API right now (and avoid adding more complexity to the current PR), is to just not expose the pixel format and use a SkImageInfo::MakeN32() regular SkiaSurface with the image that is given from the clipboard, and do the conversion in `DragDataProviderWin::getImage()` to the RGBA `os::Surface`.

dacap: I think what we are looking for is a function to convert the clip image (which can be obtained in several formats) into a standard os::SkiaSurface (N32), without exposing more color formats at the moment (because this is a more advanced feature for the laf-os library which is not required).

> Why would we do the conversion in `DragDataProviderWin`?

If you are getting a 24bpp image from macOS and Windows (which probably we do because clip tries to keep the original format) you can create a function that is called from both sides. We need a `convert_clip_image_to_surface()` (internal function/platform independent) that can be used in these getImages() implementations.

martincapello: What if, now that laf depends on clip, we replace this:
```c++
    virtual Ref<Surface> makeSurface(int width, int height, const os::SurfaceFormatData& sf, const unsigned char* data = nullptr) = 0;
```
with this
```c++
    virtual Ref<Surface> makeSurface(const clip::image_spec& spec, const clip::image& image) = 0;
```
And do the convert clip image to surface there?

dacap: Yes, we can create that function. Anyway the spec is already inside the image (`image.spec()`):

    virtual Ref<Surface> makeSurface(const clip::image& image) = 0;

A couple of things about this:

1) Forward declare `clip::image` (don't include the clip header file from `os/system.h`)
2) After merging this PR I'll add a switch to avoid using the clip library from laf (and disabling the whole drag-and-drop stuff). But this is not part of this PR now.

dacap: Yes, you can replace:
```
 Ref<Surface> makeSurface(int width, int height,
                          const os::SurfaceFormatData& sf,
                          const unsigned char* data);
```
with the new one that receives the `clip` image.

I'm not sure, but probably this makeSurface() can be defined in CommonSystem, because it doesn't depend on Skia (I guess).

os/win/dnd.cpp
martincapello: I'm trying to do this, but I can't get the DragQueryFile (didn't test getImage() yet) to work when refactored this way.

martincapello: It does compiles with some modifications, but it throws a runtime error

martincapello: I was able to make it work if:
```c++
  operator T() {
    return m_data;
  }
```
But then it complains when calling read_png from getImage()...

dacap: I think as there is a `GLock<HDROP>` we cannot change the `GLock` and just leave it as it is (receiving a pointer as `T`).

os/win/window.h
martincapello: I think so, will try it.

martincapello: Done

os/skia/skia_surface.cpp
martincapello: > I'm not sure, but probably this makeSurface() can be defined in CommonSystem, because it doesn't depend on Skia (I guess).

Actually the current makeSurface() doesn't depend on Skia either, isn't it? I would leave the method in System, but if you prefer to have it in CommonSystem, I'm fine. Just confirm which alternative should I go for.

dacap: I'm not sure what do you mean with "current makeSurface"? The makeSurface() version created in this PR (that received data and format) must be gone and replaced with the clip one.

martincapello: Got it now. I was confused about the CommonSystem class, thank you for the clarification.

os/win/dnd.cpp
```diff
@@ -54,129 +54,139 @@ gfx::Point drag_position(HWND hwnd, POINTL& pt)
...
-
-  STGMEDIUM medium;
-  medium.hGlobal = nullptr;
-  FORMATETC fmt;
-  fmt.ptd = nullptr;
-  fmt.dwAspect = DVASPECT_CONTENT;
-  fmt.lindex = -1;
-  fmt.tymed = TYMED_HGLOBAL;
 
+  auto makeSurface = [](const clip::image& img) {
```
martincapello, line 149: I know this could be a private function, but I preferred to do it here which is closer to the place it is used. Also, this is going to disappear in a future commit since os::instance()->makeSurface() will be called directly after its pending changes are implemented.

os/skia/skia_surface.cpp
martincapello: Seems that I cannot call the "generic" makeSurface() because there isn't one. The only implementation of that makeSurface is in the SkiaSystem class, then when I try to call it from CommonSystem it is actually a pure virtual function and then I cannot call it (at least as far as I understood...).

os/win/dnd.cpp
```diff
@@ -54,129 +54,139 @@ gfx::Point drag_position(HWND hwnd, POINTL& pt)
...
+    if (m_result != S_OK)
+      return nullptr;
+
+    return GLocked<T>(m_medium.hGlobal);
+  }
+
+private:
+  void release() {
+    if (m_result == S_OK) {
+      ReleaseStgMedium(&m_medium);
```
martincapello, line 109: I have to fix this...I think that when the GLocked is destroyed, after it unlocks the HGLOBAL, then that is the moment to release the medium.

dacap: OK. Just in case, I'd go back to the previous name `GLock` or renamed it to `HGlobalLock` (generally lock wrappers don't have a "locked" in its name, e.g. scoped_lock, lock_guard, etc,).

martincapello: Just as an aside: it actually made (and makes) sense to me that the thing returned is actually a locked HGLOBAL, and not a lock that I can lock or unlock. Anyway I'll return to the GLock. Thanks.

dacap: I'd highly recommend to keep consistence of naming things with current standards or other project, you can search for example in other projects (e.g. [Chrome](https://source.chromium.org/chromium/chromium/src/+/main:base/win/scoped_hglobal.h;l=19?q=hglobal%20lock&ss=chromium)).

martincapello:
This force push was to keep all the changes of the IDataObject wrapper simplification in one commit.

os/win/dnd.cpp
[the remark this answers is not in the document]
martincapello: Should I do something about this?

os/skia/skia_surface.cpp
martincapello: I've been trying to implement this in CommonSystem:
```c++
#if CLIP_ENABLE_IMAGE
  SurfaceRef makeSurface(const clip::image& image) override {
    return makeSurface(image.spec().width, image.spec().height);
  }
#endif
```
I'm just calling the "generic" makeSurface (if that is the one)...but I'm getting this compilation error:
```
C:\...\aseprite\laf\os\common\system.h(129): error C2660: 'os::CommonSystem::makeSurface': function does not take 2 arguments
  C:\...\aseprite\laf\os/common/system.h(128): note: see declaration of 'os::CommonSystem::makeSurface'
  C:\...\aseprite\laf\os/common/system.h(129): note: while trying to match the argument list '(const unsigned long, const unsigned long)'
```

Any idea?

martincapello: I've found out that the issue is because the generic makeSurface:
```c++
virtual Ref<Surface> System::makeSurface(int, int, const os::ColorSpaceRef& = nullptr) = 0;
```
Is declared abstract in `System`.

Then I guess that CommonSystem doesn't have any implementation to link to. 

So there are two options I think:
1) Implement the generic makeSurface in CommonSystem as a dummy method:
```c++
Ref<Surface> System::makeSurface(int, int, const os::ColorSpaceRef& = nullptr) { return nullptr; }
```
 And then implement the new makeSurface in CommonSystem as follows:
```c++
#if CLIP_ENABLE_IMAGE
  SurfaceRef makeSurface(const clip::image& image) override {
    return makeSurface(image.spec().width, image.spec().height);
  }
#endif
```

2) Implement the new makeSurface directly on SkiaSystem.

dacap: I'm not in the computer now, please try doing `this->makeSurface(...)`

martincapello: > I'm not in the computer now, please try doing `this->makeSurface(...)`

Double check this, it doesn't work. Same compilation error as in https://github.com/aseprite/laf/pull/83#discussion_r1569245021

martincapello:
This is ready for review. I'm going to add support for dragging URLs (https://github.com/aseprite/laf/pull/83#issuecomment-2056959996)

dacap:
Just in case I'll start reviewing this PR.

dacap:
Something about 3f13e5f867017cb73204bb53bdec1a10ea981d32 I'm not sure why the SurfaceFormat enum was refactored to a enum class. It looks like something unnecessary or at least, when we refactor an `enum` to a `enum class` it's better to remove the suffix part, e.g. instead of `kRgbaSurfaceFormat` we could use:

```c++
enum class SurfaceFormat {
  kRgba,
};
```

I'll leave this as it is in the PR and see how we finally rename these enumerations in #11 (its a task of a bigger API redesign).

martincapello:
> Something about [3f13e5f](https://github.com/aseprite/laf/commit/3f13e5f867017cb73204bb53bdec1a10ea981d32) I'm not sure why the SurfaceFormat enum was refactored to a enum class. 

Because of this comment: https://github.com/aseprite/laf/pull/83#discussion_r1532500638

I took the chance to change it since I was making changes already.

dacap:
> Because of this comment: [#83 (comment)](https://github.com/aseprite/laf/pull/83#discussion_r1532500638)
> 
> I took the chance to change it since I was making changes already.

We have 3 kind of enums:
```c++
enum Something { SomethingOne, SomethingTwo, SomethingThree };
enum Something { OneSomething, TwoSomething, ThreeSomething };
enum class Something { One, Two, Three };
```

But we don't have (at least I don't remember, or we should try to avoid) these ones:
```c++
enum class Something { OneSomething, TwoSomething, ThreeSomething };
enum class Something { SomethingOne, SomethingTwo, SomethingThree };
```
Because this force us to write `Something` two times, e.g. `Something::OneSomething`, etc. Anyway I've just rebased the `dnd` branch skipping 3f13e5f867017cb73204bb53bdec1a10ea981d32 and everything went just fine (and we can close #4414 as none of these changes are required).

dacap (changes requested):
A lot of work already done 👍 I've commented just about some minor adjustments and pushed some changes to the example. The main thing to fix is to take care of the `os::Window::scale()` in the mouse/drag position.

os/common/system.cpp
```diff
@@ -0,0 +1,186 @@
...
+    case 32:
+      get_rgba = get_rgba32;
+      break;
+    case 24:
+      get_rgba = get_rgba24;
+      break;
+    case 16:
+      get_rgba = get_rgba16;
+      break;
+  }
```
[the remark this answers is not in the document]
martincapello: Let's do this now, I wasn't sure about using std::function either here.

os/dnd.h
```diff
@@ -0,0 +1,135 @@
...
+os::SurfaceRef default_decode_jpg(const uint8_t* buf, const uint32_t len);
+os::SurfaceRef default_decode_bmp(const uint8_t* buf, const uint32_t len);
+os::SurfaceRef default_decode_gif(const uint8_t* buf, const uint32_t len);
+
+namespace os {
+
+  class Window;
+
+#if CLIP_ENABLE_IMAGE
+  using DecoderFunc = std::function<SurfaceRef(const uint8_t* buf, const uint32_t len)>;
```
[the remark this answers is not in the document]
martincapello: I was about to do this yesterday, but then I saw that std::functions are used in other places in laf, so I left it as is. Now I'm going to change it, but I would like to know if the other std::functions will be replaced in the future, or there is any other reason to not use here a std::function

os/dnd.h
```diff
@@ -0,0 +1,135 @@
...
+    // Sets what will be the outcome of dropping the dragged data when it is
+    // accepted by the target window. Only one of the enum values should be passed,
+    // do not combine values using bitwise operators.
+    void dropResult(DropOperation operation) { m_dropResult = operation; }
+    // Set this to true when the dropped data was accepted/processed by the
+    // target window, or set to false otherwise.
+    void acceptDrop(bool value) { m_acceptDrop = value; }
+
+    bool sourceSupports(DropOperation op) {
+      return static_cast<int>(m_supportedOperations) & static_cast<int>(op);
```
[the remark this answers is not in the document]
martincapello: Why is it needed? I want to return true if `static_cast<int>(m_supportedOperations) & static_cast<int>(op)` is not zero, which should be only true if the bits from `m_supportedOperations` ANDed with the bits of `op` gives a value greater than zero. Or is this just a clang-tidy thing?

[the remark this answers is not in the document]
dacap: It depends on how `sourceSupports()` should work if we specify several flags. For example I was thinking `sourceSupports(Copy | Link)` should return true if the source supports both operations (Copy+Link, so this change is required). But now I can imagine something like `sourceSupports(Any)` and I'm confused, because it should return true if any operation is supported (as it's now in the PR), but I was thinking about `All` instead of `Any` (returning true if all operation are supported).

So I'm not sure if `DropOperation::Any` should be called `DropOperation::All` or probably we should remove this constant to avoid confusion, or keep it as `Any` rename it as `All` depending on what is expected in all functions that receive/return these flags.

os/dnd.h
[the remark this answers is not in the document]
dacap: Both are right. The good part of the `std::function<>` is that it supports lambda functions with captures, and the function pointer is just a function. Probably it can be a `std::function<>` if we need it (like `handleWindowResize`) anyway I'd like to find a good solution for these kind of callbacks in the API (at the moment they are just hacks).

os/dnd.h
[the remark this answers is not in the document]
martincapello: I see. The semantics of `sourceSupports(Copy | Link)` is that it returns true if it supports Copy OR Link (or both maybe). Since you are using a bitwise OR, it actually makes sense to me to use it in this way. And the same reasoning was applied to name the DropOperation `Any`, because the function returns true if it supports any of the indicated flags (not all).

If you want to change the behavior of this function to work as you suggested, then yes, the name `All` would be appropriate in this case. Let me know what you want to do.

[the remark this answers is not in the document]
dacap: Probably I'd prefer to change the behavior with the `== static_cast<int>(op)` and rename to `All` option. We're already doing exactly this in Aseprite in `doc::Layer::hasFlags()` and `ui::Widget::hasFlags()`.

This is because a bitwise OR works like an `a || b` only if the operands are the same bit, but here we generally combine different bits/flags, in this case a bitwise OR is like an addition of the values (a+b).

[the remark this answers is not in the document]
martincapello: I know how bitwise OR works, no need to explain it, thanks anyway. As I said, the current behavior was intentionally done that way. However I'm happy to change it because it is what you prefer. Thank you.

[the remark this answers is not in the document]
dacap: Just in case I'm not lecturing you how bitwise OR works, I'm explaining the logic, or the though process we could say, behind these "hasFlags" functions I had at the moment and why I think it makes sense to works in this way.

> because it is what you prefer

In case you think this is wrong or is a better way, you can just propose that better way. For example, one alternative would be to have functions like `supportsAny()` or `supportsAll()`, which makes no mistake on the intention from the client side.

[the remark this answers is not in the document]
martincapello: I have already done what I think it is the best way. It didn't match your expectations, that's fine, there is not much else to do on my side besides doing what you expected. I'm not saying that what you say is wrong, it is just a different approach which also works.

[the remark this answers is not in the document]
martincapello: Also, I prefer to have just one function for this, it doesn't worth it to have two for almost the same.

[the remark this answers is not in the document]
dacap: > EDIT: Sorry, you was saying one or the other...it could be a possibility to name it supportsAny, it makes sense...but now that I know what you expected, I would hardly try to make it as I thought. Thank you anyway.

I just saw your edit. No problem about this, I prefer the shorter version too (`supports()`) and in most cases only one operation will be used (Copy)

dacap:
Good job @martincapello! merged just right now 🍾

martincapello:
> Good job @martincapello! merged just right now 🍾

😅 tough one

PR #90, opened 2024-05-30
Implement different approach to events queue handling

### LINKED ISSUE #89: macOS event queue won't wakeup under certain circumstances
I've found that when a laf client calls os::queue_event(ev) on macOS there is a chance that the event is not processed until some other OS event "wakes up" the queue. This can be seen when the getEvent() method is called from the main thread without a timeout, because when using a timeout, the timeout wakes up the queue and the issue is not perceived.

The issue can happen when the os::queue_event(ev) call takes place after line 77 and before line 82 in the following fragment of code:
https://github.com/aseprite/laf/blob/695bb59d1a7de36747236f3f46854b614725da77/os/osx/event_queue.mm#L77-L86 

Because the m_sleeping flag is still false in that moment, the queueEvent() method won't call the wakeUpQueue() method. Then the event got queued (in m_events queue) but then the queue will go to sleep (call to nextEventMatchingMask with untilDate=distantFuture because we are not using a timeout), waiting for any OS event to wake it up.

I think I've a fix for this which is basically a different approach, I'll try it with Aseprite and then create a PR.

EDIT: I've been trying this for a bit with Aseprite and it seems to work fine. Also I want to clarify why I need this fix:
While developing our laf based UI library I wanted to let the main thread just read all the OS events and if there is no OS events then just wait for the next one, no timeout, just sit and wait. When an OS event is available I just translate it to an app event and then push it to the app event queue. This is because in a background thread I keep reading the app queue to then send app events to the appropriate widgets. Still lots of work to do, but I just wanted to let you know which was the use case made the current implementation fail.

martincapello:
Fix #89

### DIFF
--- a/os/osx/event_queue.h
+++ b/os/osx/event_queue.h
@@ -9,11 +9,9 @@
 #define OS_OSX_EVENT_QUEUE_INCLUDED
 #pragma once
 
+#include "base/concurrent_queue.h"
 #include "os/event.h"
 #include "os/event_queue.h"
-
-#include <queue>
-#include <mutex>
 
 namespace os {
 
@@ -26,8 +24,7 @@ public:
   void clearEvents() override;
 
 private:
-  mutable std::mutex m_mutex;
-  std::deque<os::Event> m_events;
+  base::concurrent_queue<Event> m_events;
 };
 
 using EventQueueImpl = EventQueueOSX;
--- a/os/osx/event_queue.h
+++ b/os/osx/event_queue.h
@@ -1,5 +1,5 @@
 // LAF OS Library
-// Copyright (C) 2018-2021  Igara Studio S.A.
+// Copyright (C) 2018-2024  Igara Studio S.A.
 // Copyright (C) 2015-2016  David Capello
 //
 // This file is released under the terms of the MIT license.
@@ -9,11 +9,12 @@
 #define OS_OSX_EVENT_QUEUE_INCLUDED
 #pragma once
 
-#include "base/concurrent_queue.h"
 #include "os/event.h"
 #include "os/event_queue.h"
 
-#include <atomic>
+#include <map>
+#include <mutex>
+#include <cstdint>
 
 namespace os {
 
@@ -26,10 +27,11 @@ public:
   void clearEvents() override;
 
 private:
-  void wakeUpQueue();
+  void extractEvent(Event& ev, uint32_t eventId);
 
-  base::concurrent_queue<Event> m_events;
-  std::atomic<bool> m_sleeping;
+  mutable std::mutex m_mutex;
+  std::map<uint32_t, os::Event> m_events;
+  uint32_t m_nextEventId;
 };
 
 using EventQueueImpl = EventQueueOSX;
--- a/os/osx/event_queue.h
+++ b/os/osx/event_queue.h
@@ -12,9 +12,8 @@
 #include "os/event.h"
 #include "os/event_queue.h"
 
-#include <map>
+#include <queue>
 #include <mutex>
-#include <cstdint>
 
 namespace os {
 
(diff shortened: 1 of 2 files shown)


dacap:
I didn't test or review this, but just at a glance I'd like to ask if it's really needed a map? why not just a sequential vector of events? (as events must be always processed in order, but maybe I'm missing something).

Anyway seeing the original issue, I'll try to reproduce the error and see if there is another alternative (e.g. if the concurrent_queue could offer an alternative try_pop() function returning the lock to its internal mutex and we can unlock it from the client side when the if-statement ends).

dacap:
> I'll try to reproduce the error and see if [...]

Just as a side note this will be next week, but you can give a try to an alternative implementation doing this of the "try_pop_with_lock" thing.

martincapello:
> I didn't test or review this, but just at a glance I'd like to ask if it's really needed a map? why not just a sequential vector of events? (as events must be always processed in order, but maybe I'm missing something).

No, it is not really needed, just something where we can store the events. I've used a map because I thought that giving the events an id and been able to find it by key was a good idea. But now I think that we could use a regular std::deque, and avoid the use of ids. I will try this and update the PR.

> 
> Anyway seeing the original issue, I'll try to reproduce the error and see if there is another alternative (e.g. if the concurrent_queue could offer an alternative try_pop() function returning the lock to its internal mutex and we can unlock it from the client side when the if-statement ends).

Okay. I just need to make that use case work. To reproduce the issue more easily yo can add a sleep(2ms) before line 82.

martincapello:
> > I'll try to reproduce the error and see if [...]
> 
> Just as a side note this will be next week, but you can give a try to an alternative implementation doing this of the "try_pop_with_lock" thing.

Not sure why should we look for another alternative if we have one that seems to work already. Also I actually like this implementation 👉 👈 ...so I hope you understand that it is difficult for me to replace it with anything else, because I put some thought, time, work, and love in this. Only if proven that this doesn't work I would be able to look for a better alternative. If you came up fix a different fix, that's fine by me, I won't go mad nor anything.

dacap:
No problem, I'll try to review this next week in detail and see what's wrong with the current impl and what offer this fix.

The issue here that makes me noise is:
1. handling two collections of events and moving between these (concurrent_queue <-> map/dequeue) instead of using just the concurrent queue for a locking/threading issue (like in other platforms).
2. handling events is the most performant thing that must happen (as is the first thing to process each mouse event)
3. there is a comment removed about "changing the display color profile", not sure if that was tested, probably this code is not required anymore

martincapello:
> No problem, I'll try to review this next week in detail and see what's wrong with the current impl and what offer this fix.
> 
> The issue here that makes me noise is:
> 
> 1. handling two collections of events and moving between these (concurrent_queue <-> map/dequeue) instead of using just the concurrent queue for a locking/threading issue (like in other platforms).

Oh there is no more concurrent queue. It was replaced, so only one collection.

> 2. handling events is the most performant thing that must happen (as is the first thing to process each mouse event)

I agree!

> 3. there is a comment removed about "changing the display color profile", not sure if that was tested, probably this code is not required anymore

Will take a look at this.

dacap:
> Oh there is no more concurrent queue. It was replaced, so only one collection.

Oh ok, check that base::concurrent_queue<> is just what you have now: a mutex + deque.

martincapello:
> Oh ok, check that base::concurrent_queue<> is just what you have now: a mutex + deque.

You are right, the original idea of having the mutex in the class was because I wanted to lock several steps, not only the queue, because I was trying to generate ids to keep track of the events. Since that is not actually useful, I think I can just go back to using the concurrent queue. Will try it. Thank you.

martincapello:
I've been reviewing this solution and one thing that annoys me is that it effectively duplicates the events in the OS queue, I mean, each OS event is pushed to the concurrent queue and then is pushed again to the OS event as an application defined event just to signal that an event should be popped from the concurrent queue. So I went back to the original code and tried a different fix.

PR #103, opened 2024-08-19
Improve skia text shaping & bounds calculation

ckaiser:
Fixes Skia crashes on Windows by using different bounds calculations and a more complete shape() function call.

### DIFF
--- a/text/skia_font.cpp
+++ b/text/skia_font.cpp
@@ -71,11 +71,13 @@ int SkiaFont::height() const
 
 int SkiaFont::textLength(const std::string& str) const
 {
-  return m_skFont.measureText(
-    str.c_str(),
-    str.size(),
-    SkTextEncoding::kUTF8,
-    nullptr);
+  return std::ceil(
+    m_skFont.measureText(
+      str.c_str(),
+      str.size(),
+      SkTextEncoding::kUTF8,
+      nullptr)
+  );
 }
 
 float SkiaFont::measureText(const std::string& str,
--- a/text/skia_font.cpp
+++ b/text/skia_font.cpp
@@ -124,9 +124,8 @@ glyph_t SkiaFont::codePointToGlyph(codepoint_t codepoint) const
 
 gfx::RectF SkiaFont::getGlyphBounds(glyph_t glyph) const
 {
-  float widths;
   SkRect bounds;
-  m_skFont.getWidthsBounds(&glyph, 1, &widths, &bounds, nullptr);
+  m_skFont.getWidthsBounds(&glyph, 1, nullptr, &bounds, nullptr);
   return os::from_skia(bounds);
 }
 
--- a/text/skia_text_blob_shaper.cpp
+++ b/text/skia_text_blob_shaper.cpp
@@ -34,7 +34,8 @@ public:
                    const gfx::PointF& offset,
                    TextBlob::RunHandler* subHandler)
     : m_builder(utf8Text, os::to_skia(offset))
-    , m_subHandler(subHandler) { }
+    , m_subHandler(subHandler)
+    , m_buffer() { }
 
   sk_sp<SkTextBlob> makeBlob() {
     return m_builder.makeBlob();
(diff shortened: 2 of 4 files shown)


text/skia_text_blob_shaper.cpp
```diff
@@ -131,17 +132,33 @@ TextBlobRef SkiaTextBlob::MakeWithShaper(
...
+
+    shaper->shape(text.c_str(),
+                  text.size(),
+                  *fontRun,
+                  *bidiRun,
+                  *scriptRun,
+                  *languageRun,
+                  std::numeric_limits<float>::max(),
+                  &shaperHandler);
+
```
dacap, line 161: It looks like `MakeBiDiRunIterator` and `MakeScriptRunIterator` functions are legacy too.

Probably we could do something similar to `SkShaperPrimitive::shape()` (using values instead of smart pointers for each run iterator):
```c++
void SkShaperPrimitive::shape(const char* utf8,
                              size_t utf8Bytes,
                              const SkFont& font,
                              bool leftToRight,
                              SkScalar width,
                              RunHandler* handler) const {
    std::unique_ptr<FontRunIterator> fontRuns(
            MakeFontMgrRunIterator(utf8, utf8Bytes, font, nullptr));
    if (!fontRuns) {
        return;
    }
    // bidi, script, and lang are all unused so we can construct them with empty data.
    TrivialBiDiRunIterator bidi{0, 0};
    TrivialScriptRunIterator script{0, 0};
    TrivialLanguageRunIterator lang{nullptr, 0};
    return this->shape(utf8, utf8Bytes, *fontRuns, bidi, script, lang, nullptr, 0, width, handler);
}
```
Anyway I tried to replace with these iterators and didn't work (I got a crash), so there is something I'm missing here.

text/skia_text_blob_shaper.cpp
```diff
@@ -34,7 +34,8 @@ class ShaperRunHandler final : public SkShaper::RunHandler {
                    const gfx::PointF& offset,
                    TextBlob::RunHandler* subHandler)
     : m_builder(utf8Text, os::to_skia(offset))
-    , m_subHandler(subHandler) { }
+    , m_subHandler(subHandler)
+    , m_buffer() { }
```
dacap, line 38: The `m_buffer` is a `Buffer` struct without constructor and pointers, is this the line that fixed the crash? (with this change we're initializing `m_buffer`'s internal pointers to `nullptr`)

text/skia_font.cpp
```diff
@@ -71,11 +71,13 @@ int SkiaFont::height() const
...
-    str.size(),
-    SkTextEncoding::kUTF8,
-    nullptr);
+  return std::ceil(
+    m_skFont.measureText(
+      str.c_str(),
+      str.size(),
+      SkTextEncoding::kUTF8,
+      nullptr)
+  );
```
dacap, line 80: I think this is the best to do at the moment, in a near future `textLength()` should return a `float`.

text/text_blob.cpp
```diff
@@ -69,10 +69,12 @@ gfx::RectF TextBlob::RunInfo::getGlyphBounds(const size_t i) const
   if (bounds.isEmpty()) {
     FontMetrics metrics;
     font->metrics(&metrics);
-    bounds.w = metrics.avgCharWidth;
-    bounds.h = -metrics.capHeight;
+    // avgCharWidth can be 0, so we grab the next most useful thing, the height
+    bounds.w = metrics.avgCharWidth > 0 ? metrics.avgCharWidth : metrics.xHeight;
+    bounds.h = 1.0;
```
dacap, line 74: I'm not sure about this one, `xHeight` is the height of the lowercase `x` character. It feels wrong to use this value as the whitespace width (and hard coding 1.0 as the height).

To test this we can use the `text_shape` which is a mini text editor, put the cursor at a whitespace to test the whitespace bounds.

We should be able to set bounds.w = the advance in X axis for whitespaces, but I'm not sure where to find this value on Skia.

dacap (changes requested):
Hi @ckaiser, some changes to be made/test. You can test with text_shape example the whitespace bounds, we should find a way to fill that glyph bounds correctly.

text/skia_text_blob_shaper.cpp
ckaiser: No, this was just a compiler suggestion and I tend to prefer initializing everything, but I don't think it changes anything

text/text_blob.cpp
ckaiser: Yeah this one's the most contentious change, the hardcoded 1 was to try and fix a specific bug of things being too tall in some circumstances, it should not stay like this. I need to look into more of the documentation

text/skia_text_blob_shaper.cpp
ckaiser: I'll look into it, it's very strange that we're getting different behaviors per-platform

text/skia_text_blob_shaper.cpp
dacap: Actually this initialization does initialize the `m_buffer` members to `nullptr`, I've tested that: https://godbolt.org/z/9s5YjETWP

Without this the m_buffer member pointers are not set to `nullptr` (that's because `Buffer` struct does't have a constructor).

text/skia_text_blob_shaper.cpp
dacap: Oh in this case your code worked on Linux :+1: I was just commenting this because it might be better if we avoid calling those legacy `Make*RunIterator` functions and create those iterators in the stack directly.

text/text_blob.cpp
ckaiser: I'm trying a new approach with just the normal " " character glyph bounds and it looks to behave correctly - at least more consistently than attempting to use FontMetrics. I think the ideal solution would be to shape the entire set of glyphs in one go with skia using `getWidthsBounds` but that'd mean a full refactor of how the runs are working currently.

text/skia_text_blob_shaper.cpp
ckaiser: The thing is that we need to use these functions because they have different implementations depending on what's available, so on a platform without harfbuzz or on a compile without it, it'll use a trivial one - etc. If we don't use them we'd have to include that logic ourselves.

The legacy stuff is kind of interesting too because the only place where that `SK_DISABLE_LEGACY_SKSHAPER_FUNCTIONS` is declared is in the Bazel compile scripts, not in the GN ones, so it's not even possible to disable these functions for anyone building aseprite atm. I think it's probably ok to keep using them for now.

dacap:
Just merged :+1: with a couple of extra fixes: 583d045dba4a5218b3732e44607423ca657c383c and 06279da54f2849ee042897af5457efa442401ab4

PR #105, opened 2024-09-05
[x11] Add support for tablet device detection

exLeute:
Fix tablet detection system by checking for device axes instead of string comparison.

### DIFF
--- a/os/x11/xinput.cpp
+++ b/os/x11/xinput.cpp
@@ -77,75 +77,52 @@ void XInput::load(::Display* display)
   if (!system)
     return;
 
-  std::string userDefinedTablet = system->tabletOptions().detectStylusPattern;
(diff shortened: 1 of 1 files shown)


exLeute:
@dacap sorry for late pr, don't really have time, but main logic should be ready, also i think i have not added general settings to manually put tablet, but it still should work better than detecting tablet by string, because i could not debug this functionality. But main detection thing works on both of my x11 systems as intended.

dacap:
Hi @exLeute, not sure if there are other devices with more than 3 axis that are not a stylus (e.g. a mouse with horizontal scroll). So not sure if this would work for all cases.

exLeute:
True @dacap, but it can be polished and added as an optional detection method, or something like drop down list of all "xinput" devices

PR #109, opened 2024-10-09
Add possibility to set a webp decoding function

martincapello:
This is useful in Windows to let the application set this function to decode webp data present when some applications provide it when dropping some element (like Chrome does when dragging&dropping an image from a result page).


dacap:
@martincapello just merged, remember to create the bug/feature branches in your own fork (not in aseprite/laf or igarastudio/laf).

PR #115, opened 2024-11-28
Don't add new ref to the System instance which is being destroyed (fix aseprite/aseprite#4811)

dacap:
Regression introduced in 3bea53125ccafce878d17fcf8d9d192ad69fafb4, when os::instance() was refactored to os::System::instance() to return a SystemRef instead of a System*, and ~WindowWin() destructor asks for the current system() (and WindowWin::system() is trying to create a new System reference, but the System is being destroyed as it was completely unreferred, i.e. ref=0.)

### DIFF
--- a/os/common/system.cpp
+++ b/os/common/system.cpp
@@ -25,14 +25,22 @@
   #include "clip/clip.h"
 #endif
 
+#include "base/debug.h"
+
 namespace os {
 
 // Weak reference to the unique system instance. This is destroyed by
 // the user (with the main SystemRef to the system).
 System* g_instance = nullptr;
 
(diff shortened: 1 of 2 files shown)


dacap:
Just in case, this can be tested with the new `view_benchmark.exe` benchmark: https://github.com/aseprite/aseprite/blob/beta/src/ui/view_benchmark.cpp (the benchmark crashes at the end without this patch).

dacap:
After some @martincapello testing, other issues were found but we weren't able to reproduce the bug again. I'll merge this PR as it is as at least is a real mitigation of Sentry crash reports from #4811.

PR #116, opened 2024-12-03
Add modal window treatment to macOS

Gasparoken:
This fix is the first step to solve issue aseprite/aseprite#4561

### DIFF
--- a/os/osx/window.mm
+++ b/os/osx/window.mm
@@ -1,5 +1,5 @@
 // LAF OS Library
-// Copyright (C) 2019-2021  Igara Studio S.A.
+// Copyright (C) 2019-2024  Igara Studio S.A.
 // Copyright (C) 2012-2017  David Capello
 //
 // This file is released under the terms of the MIT license.
@@ -121,6 +121,9 @@
       self.level = NSFloatingWindowLevel;
       self.hidesOnDeactivate = true;
     }
+
+    if (spec->modal())
+      self.level = NSModalPanelWindowLevel;
 
     // Hide the "View > Show Tab Bar" menu item
     if ([self respondsToSelector:@selector(setTabbingMode:)])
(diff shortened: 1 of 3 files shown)


os/osx/window.mm
```diff
@@ -120,6 +120,8 @@ - (WindowOSXObjc*)initWithImpl:(os::WindowOSX*)impl
     if (spec->floating()) {
       self.level = NSFloatingWindowLevel;
       self.hidesOnDeactivate = true;
+      if (spec->modal())
+        [self setLevel:NSModalPanelWindowLevel];
     }
```
dacap, line 125: We're already setting the level property if it's floating, probably we can have a modal / not floating (this is not the case of Aseprite which sets the "floating" property for all non-main windows):
```suggestion
    if (spec->floating()) {
      self.level = NSFloatingWindowLevel;
      self.hidesOnDeactivate = true;
    }
    if (spec->modal())
      self.level = NSModalPanelWindowLevel;
```

dacap (changes requested):

os/osx/window.mm
```diff
@@ -122,6 +122,9 @@ - (WindowOSXObjc*)initWithImpl:(os::WindowOSX*)impl
       self.hidesOnDeactivate = true;
     }
 
+    if (spec->modal())
+      [self setLevel:NSModalPanelWindowLevel];
```
dacap, line 126:
```suggestion
      self.level = NSModalPanelWindowLevel;
```

dacap:
I'll merge this, it would be nice to block the whole hierarchy of parent's floating windows when a modal is displayed, but at the moment this works better than the current `main` to fix the bug.

Gasparoken:
> I'll merge this, it would be nice to block the whole hierarchy of parent's floating windows when a modal is displayed, but at the moment this works better than the current `main` to fix the bug.

Great! Remember to approve https://github.com/aseprite/aseprite/pull/4835 so that issue https://github.com/aseprite/aseprite/issues/4561 is resolved in Aseprite.

PR #117, opened 2024-12-05
Add a string_view variant of base::split_string()

ckaiser:
Adds a string_view version of `base::split_string`

Benchmarked difference between the two versions of the function puts the string_view version at around 36-37% faster, although it wasn't extensive testing, just a little microbenchmark:

```cpp
  auto t0 = std::chrono::high_resolution_clock::now();
  {
    const std::string_view string = "Hello,World";
    for (int i = 0; i < 200000; i++) {
      std::vector<std::string_view> result;
      base::split_string("Hello,World", result, ",");
    }
  }

  auto t1 = std::chrono::high_resolution_clock::now();
  auto view = std::chrono::duration_cast<std::chrono::microseconds>(t1 - t0).count();
  t0 = std::chrono::high_resolution_clock::now();

  {
    const std::string string2 = "Hello,World";
    for (int i = 0; i < 200000; i++) {
      std::vector<std::string> result;
      base::split_string(string2, result, ",");
    }
  }

  t1 = std::chrono::high_resolution_clock::now();
  auto noView = std::chrono::duration_cast<std::chrono::microseconds>(t1 - t0).count();

  std::cout << "string_view: " << view << " us" << std::endl;
  std::cout << "string:      " << noView << " us" << std::endl;
```

### DIFF
--- a/base/split_string.cpp
+++ b/base/split_string.cpp
@@ -13,12 +13,10 @@
 #include <algorithm>
 
 namespace {
-
   struct is_separator {
     const std::string* separators;
 
-    is_separator(const std::string* seps) : separators(seps) {
-    }
+    is_separator(const std::string* seps) : separators(seps) {}
 
     bool operator()(std::string::value_type chr) const
     {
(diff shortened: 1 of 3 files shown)


dacap:
Hi @ckaiser, we were avoiding `std::string_view` at the moment but just for fear of macOS 10.9 incompatibilities. The GitHub CI worked (it compiles with `CMAKE_OSX_DEPLOYMENT_TARGET=10.9` for macOS in the same way we compile the official distribution) so it should be safe to start using `std::string_view` (there are a lot of TODO/places where we use `substr()` and it'd be nice to use string_views directly).

About `base::split_string`, just in case there exists `base::tok()` function (in `base/tok.h`) which is a little faster and have some kind of different API if required.

The `base::tok()` function is based on https://github.com/dacap/tok, and there you can find some benchmarks too: https://github.com/dacap/tok/blob/main/benchmarks/tok_benchmarks.cpp
(I'd like to benchmark `base::tok()` using `string_view` some time in the future).

PR #119, opened 2024-12-16
Add base::get_absolute_path_from_relative() function

Gasparoken:
Part of https://github.com/aseprite/aseprite/issues/4851

It's wanted to reconstruct the absolute path starting from the relative/absolute filename in the filename entry of 'Export As' dialog.

### DIFF
--- a/base/fs.cpp
+++ b/base/fs.cpp
@@ -216,6 +216,45 @@ std::string get_relative_path(const std::string& filename, const std::string& ba
   return relativePath;
 }
 
+std::string get_absolute_path_from_relative(const std::string& base_path,
+                                            const std::string& relative_filename)
+{
+  std::vector<std::string> absoluteParts;
+  split_string(base_path, absoluteParts, path_separators);
+
+  std::vector<std::string> relativeParts;
+  split_string(relative_filename, relativeParts, path_separators);
+
+  // If the first element is empty, it means that the relative
+  // file name starts with a separator and it also means that
+  // it's an absolute address.
+  if (relativeParts.size() >= 1 && relativeParts[0] == "")
+    return relative_filename;
+
+  bool relative_fn_browsable = false;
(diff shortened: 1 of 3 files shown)


dacap:
Hi @Gasparoken, just in case, is possible that `base::get_absolute_path()` does the job? (it removes ".." elements, etc.)

Gasparoken:
> Hi @Gasparoken, just in case, is possible that `base::get_absolute_path()` does the job? (it removes ".." elements, etc.)

`base::get_absolute_path()` uses the current folder instead of an arbitrary `base_path`. That's why I implemented this new function. But maybe `base::get_absolute_path()` is enough and I just need to change the approach in Aseprite (particularly on `std::string ExportFileWindow::outputFilenameValue() const`). Let me investigate.

dacap:
> `base::get_absolute_path()` uses the current folder instead of an arbitrary `base_path`

It should work as a path normalizer if the given path is an absolute path; you can check out the tests:

https://github.com/aseprite/laf/blob/c84b89b4826b64b05a176eb15725ba526495c6a1/base/fs_tests.cpp#L246-L250

So you should be able to use something like `base::get_absolute_path(base::join_path(base_path, relative_path))`. Other way is using `base::normalize_path()` directly.

Gasparoken:
That's great! However, I found that `base::get_absolute_path()` can only handle one "../", if there are two or more "../", the function does not get to the correct folder level. Therefore, the following line fails:
```cpp
EXPECT_EQ("/user/file", get_absolute_path("/user/name/path/../../file"));
```
I'll adopt this function to resolve https://github.com/aseprite/aseprite/issues/4851, but I will need to fix the indicated bug first.

Gasparoken:
I'm closing this and will take the `base::get_absolute_path()` approach instead.

PR #121, opened 2024-12-18
[x11] Fix drag & drop receiving incorrect file with some file managers

56789a1987:
Fix aseprite/aseprite#3737

The issue is reproducible on KDE x11, Dolphin file manager. Drag & Drop keeps opening the same file.
Passing the time stamp from the `XdndDrop` event to `XConvertSelection` fixes the issue.

This can be useful as reusing the same window to open documents is still not supported on Linux.


dacap:
Thanks for this fix @56789a1987, checking the XDND specs you are right:

https://www.freedesktop.org/wiki/Specifications/XDND/#xdnddrop

I'll merge this.

PR #122, opened 2024-12-19
[X11] Fix drag & drop may stop receiving files in xwayland

56789a1987:
Fix 2 drag & drop issues observed on KDE Wayland, using XWayland.

Changed `XdndFinished` event to be sent to the source window.
If send it to the root window, after dropping the first file, subsequent files can't be dropped.

Added handling `XdndEnter` event to check data type.
If the dragged item isn't a file, then reject dropping. This gives visual cursor feedback and prevents the drop event.
Possible fix for: When an unsupported data type is dropped to the window, subsequent files can't be dropped.

Feel free to pick only the first commit. The first issue is more problematic, and the fix doesn't require too many changes.

### DIFF
--- a/os/x11/window.cpp
+++ b/os/x11/window.cpp
@@ -93,13 +93,16 @@ Atom _NET_WM_ALLOWED_ACTIONS = 0;
 
 // Atoms used for the XDND protocol
 Atom XdndAware = 0;
+Atom XdndEnter = 0;
 Atom XdndPosition = 0;
 Atom XdndStatus = 0;
 Atom XdndActionCopy = 0;
 Atom XdndDrop = 0;
+Atom XdndTypeList = 0;
 Atom XdndFinished = 0;
 Atom XdndSelection = 0;
 Atom URI_LIST = 0;
+Atom g_dndType = 0; // Accepted data type
 ::Window g_dndSource = 0;
 gfx::Point g_dndPosition;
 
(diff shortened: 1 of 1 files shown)


dacap:
Hi @56789a1987, thanks for these patches, sorry that I cannot test this (I don't have a KDE desktop at hand), but just in case, the beta branch contains several changes in the DnD implementation, could you please test that branch? (and probably patch/target this PR to the beta):

https://github.com/aseprite/laf/tree/beta
https://github.com/aseprite/laf/commit/284572b994f5cc6fc95eb96840c9e2abf00be9b6

56789a1987:
> the beta branch contains several changes in the DnD implementation, could you please test that branch? (and probably patch/target this PR to the beta)

Tested the beta branch with `examples/drag_and_drop`, generally works well.
`XdndEnter` seems have been implemented, so the second commit here can be ignored for now.

Applying #121 fixes receiving the same file issue.
And cherry picking the 1861e3f34d187124337845365dd56e1ce94408d2 fixes XWayland not accepting files issue.

Other issues: If an unsupported type (such as text) is dropped, the drop zone remains highlighted.

56789a1987:
Hi @dacap, I have tested the beta branch.
The only patches I's like to make is #121 and 1861e3f34d187124337845365dd56e1ce94408d2 mentioned above, for both main and beta branch.

6ac753a0ed955b49fcaa84f697fcc9bab3f0810b is implemented on beta, it can be ignored on both branches.

dacap:
Thanks @56789a1987 for the patch, I've just included 1861e3f34d187124337845365dd56e1ce94408d2 in `main` and I'll merge `main` to `beta` so the beta branch includes these fixes.

PR #123, opened 2024-12-26
Move font_paths files from aseprite to laf

martincapello:
Moves the font_path* files from Aseprite source into LAF's "os" namespace. I did this work because when I was trying to find the cause of https://github.com/aseprite/aseprite/issues/4795 I wanted to create an isolated simple app showing just a window with a menu using the aseprite ui, without using any aseprite code. And these files were one of the reasons that made it impossible to do it, hence they had to be moved.

### DIFF
--- a/os/CMakeLists.txt
+++ b/os/CMakeLists.txt
@@ -10,6 +10,7 @@ set(LAF_OS_SOURCES
   common/main.cpp
   common/system.cpp
   dnd.cpp
+  font_path.cpp
   system.cpp
   window.cpp)
 if(WIN32)
@@ -17,6 +18,7 @@ if(WIN32)
     win/color_space.cpp
     win/dnd.cpp
     win/event_queue.cpp
+    win/font_path_win.cpp
     win/keys.cpp
     win/native_dialogs.cpp
     win/system.cpp
@@ -34,6 +36,7 @@ elseif(APPLE)
       osx/dnd.mm
       osx/color_space.mm
       osx/event_queue.mm
+      osx/font_path_osx.mm
       osx/keys.mm
       osx/logger.mm
       osx/menus.mm
@@ -46,6 +49,7 @@ elseif(APPLE)
 else()
   list(APPEND LAF_OS_SOURCES
     x11/event_queue.cpp
+    x11/font_path_unix.cpp
     x11/keys.cpp
     x11/native_dialogs.cpp
     x11/system.cpp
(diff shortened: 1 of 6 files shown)


dacap:
Just in case, any font/text related refactor should go to beta, because there is a new laf-text library and the main branch is deprecated for this kind of stuff. Probably this can be done in the beta branch, but I'm not sure if it's needed there (you should try to do the example app with the beta branch).

martincapello:
I actually needed to use the main branch because the fix is planned for the next release of the main branch. But sure, these changes can go in beta branch. I will first see if they are still needed for beta.

martincapello:
Seems that in beta, in lines:

https://github.com/aseprite/aseprite/blob/7fb8cfc8fab97c9e4caa3a437501e4dfc4be47f3/src/app/ui/skin/skin_theme.cpp#L215-L217
```c++
      fontFilename = app::find_font(xmlDir, platformFileStr);
    if (fileStr && fontFilename.empty())
      fontFilename = app::find_font(xmlDir, fileStr);
``` 
The app::find_font() function is still being used as in the main branch. Which means that if I want to create an example with a custom Theme using just the ui module I cannot use the find_font function because it is part of Aseprite. So, I believe that the refactor still applies for beta. Will create a new PR.

dacap:
Probably we can move the find_font() function to the ui-lib in that case.

In other case, we could refactor the functionality to something inside FontMgr implementation.

dacap:
Sorry, but in case you want to use a custom font using the ui module, I think the current FontMgr can solve that: https://github.com/aseprite/laf/blob/20d13cf0c762189650c7860bd3302a0a66cfa346/text/font_mgr.h#L27 or listing the font faces of the system.

martincapello:
Okay, so, does the current FontMgr have a function that returns all the system directories that are used to look for font files?

martincapello:
Or a function to get all the font files available on the system?

dacap:
> Or a function to get all the font files available on the system?

Yes, check the `listfonts` example, the `FontMgr` is the new API to get all system font/load a specific TTF font (e.g. the Text tool / font selector are using this new API).

We want to deprecate and finally remove the `app::find_font()` in a future.

martincapello:
> We want to deprecate and finally remove the `app::find_font()` in a future.

I see, then does this means that we will change how the fonts.xml works in Aseprite? So, instead of naming the font file, we would just name the family name?

martincapello:
I'm closing this because it doesn't make sense any more, since the function will be deprecated. However, I'm still curious about the future of the processing of fonts.xml.

PR #113, opened 2024-11-21
Fix floating window disappears when another app gains focus

Gasparoken:
This is the solution for issue aseprite/aseprite#4265

PR #124, opened 2025-01-02
[osx] Mimic Windows OS's floating windows behavior

martincapello:
This PR is based on the work of PR #113 and fixes the same issues: 
Fix aseprite/aseprite#4265
Fix aseprite/aseprite#4774

The difference between both PRs is on how they locate the floating windows to manipulate some of their properties.


Gasparoken:
I just tested it and it works fine on MacOS Sequoia 15.2.

dacap:
Just tested and it works great, has the same flicker problem than the other PR but I think it might be impossible to fix. I'm going to merge it.

PR #125, opened 2025-01-04
Suggest git submodule init / update in README

martinemde:
When I went to build this (after building skia) I had these errors that blocked the build:

```
CMake Error at CMakeLists.txt:62 (add_subdirectory):
  The source directory

   path/to/laf/clip

  does not contain a CMakeLists.txt file.


CMake Error at third_party/CMakeLists.txt:10 (add_subdirectory):
  The source directory

   path/to/laf/third_party/googletest

  does not contain a CMakeLists.txt file.
```

The solution is to init and update submodules after cloning.

```
git submodule update --init
```

This will fetch the correct files and allow you to build.

It would be nice to make sure others know to init and update submodules when they fetch the code. I rarely work in projects that have submodules so I didn't immediately realize this was simple to solve.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,13 @@ need for Pixman at all).
 
 ## Compile
 
+On a fresh clone of the repository, ensure git submodules are updated.
+
+```
+git submodule init
+git submodule update
+```
+
 To compile *laf* with Skia as backend you have to specify some
 variables pointing to a valid compiled version of Skia in your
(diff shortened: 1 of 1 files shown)


dacap:
Thanks for this @martinemde, I've updated the instructions to use `git clone --recursive` which is the preferred method to initialize the repository.

PR #126, opened 2025-01-10
Add missing <cstdint> header to use uint8_t

bjorn:
This was needed to fix compilation against MinGW.

---


dacap:
Thanks @bjorn! :+1:

PR #127, opened 2025-01-11
Handle screen DPI changes on Windows

### LINKED ISSUE #4495: Native title bar is smaller on 4k display
I use Aseprite with my 4k drawing tablet, at 200% scaling (in Windows settings) but I also have two other 1080p monitors at 100% scaling. The native title bar on Aseprite looks fine on my 1080p monitors but is notably smaller on my 4k monitor. I've tried changing the Screen and UI Elements scaling options in Aseprite but neither changed the title bar's size. This issue doesn't affect any other apps I use on my 4k monitor.

Here is what Aseprite looks like on the 4k monitor:
image

* Aseprite version: 1.3.7-x64
* System: Windows 11 Pro 23H2

lampysprites:
Issue aseprite/aseprite#4606, also possibly avoids having aseprite/aseprite#2740, and I think fixes aseprite/aseprite#4495.

On several displays with diffferent scale dragging the window from one screen to another makes everything either tiny or huge, depending on the direction it's dragged. This can be an issue when the screen with higher resolution is smaller in size, like a laptop with external monitor, or a pc with a screen tablet.

I tested these changes on 100%/200% and 150%/300% screen combination, windowed/maximized window, menu and preview windows, moving window with mouse and shift+wid+arrow key, and changing screen scale in settings w/ app running. As far as I can see it behaves same as other applications (namely, notepad) now. Except, rounding of 150% scale and 1x scale setting to 1x resolution feelt a bit confusing (personally) so i rounded it up.

PR aseprite/aseprite#4934 is needed to handle these changes properly.

### Window decorations

- Laf's WinAPI constructor (laf/os/win/winapi.cpp) tries to set process' dpi awareness to PER_MONITOR_AWARE_V2, and then and to PER_MONITOR_AWARE if first one fails. In Aseprite however both actually fail with "access denied", because it can only be set once, and `src/main/settings.manifest` already set it to V1.
- In addition to enabling DPI awareness V1, window needs to call EnableNonClientDpiScaling() in create event to make titlebar resize.
- GetSystemMetric and AdjustWindowRectEx need to be replaced with dpi-aware variants when available.

Corresponding commit to Aseprite repo changes manifest file to enable V2 dpi awareness, although when it's not available V1 works identical to it. Also theorethically this can be applied without the next commit if it's too messy.

### Window content

I think it's important that window content is intact when changing screen. Think of moving a preview window - doing additional zoom/pan would be annoying. It means we need to:
- Scale window borders proportional to dpi change.
- Adjust display scale factor, not UI scale.
- Not clamp scale factor at 4x when drawing.

The easiest wai I think is to scale pixel coordinates in laf window as `scale = baseScale * displayDPI / defaultDPI`, where baseScale is now the scaling chosen by user in the settings. Scale is used for input and painting same as before, while baseScale replaces it in window settings and initial values. This however breaks backwards compatibility, and I can't think of a simple way to keep it (e.g. need to handle window sizing outside window class for that).

There's also a few ways to sprovie API for this - i picked one with fewer changes. And maybe Window::setScale should now be setBaseScale but I don't think i got naming right anyway.

Corresponding commit to Aseprite repo replaces some scale() calls with baseScale() when needed.

-----

### DIFF
--- a/os/win/window.cpp
+++ b/os/win/window.cpp
@@ -1787,8 +1787,16 @@ LRESULT WindowWin::wndProc(UINT msg, WPARAM wparam, LPARAM lparam)
           // scrollbars. In this way they are not visible, but we still
           // get their messages.
           NCCALCSIZE_PARAMS* cs = reinterpret_cast<NCCALCSIZE_PARAMS*>(lparam);
-          cs->rgrc[0].right += GetSystemMetrics(SM_CYVSCROLL);
-          cs->rgrc[0].bottom += GetSystemMetrics(SM_CYHSCROLL);
+
+          auto& winApi = system()->winApi();
+          if (winApi.GetWindowDpiAwarenessContext) {
+            UINT dpi = winApi.GetDpiForWindow(m_hwnd);
+            cs->rgrc[0].right += winApi.GetSystemMetricsForDpi(SM_CYVSCROLL, dpi);
+            cs->rgrc[0].bottom += winApi.GetSystemMetricsForDpi(SM_CYHSCROLL, dpi);
+          } else {
+            cs->rgrc[0].right += GetSystemMetrics(SM_CYVSCROLL);
+            cs->rgrc[0].bottom += GetSystemMetrics(SM_CYHSCROLL);
+          }
         }
       }
       break;
@@ -2490,10 +2498,21 @@ HWND WindowWin::createHwnd(WindowWin* self, const WindowSpec& spec)
     rc.w = spec.contentRect().w;
     rc.h = spec.contentRect().h;
     RECT ncrc = { 0, 0, rc.w, rc.h };
-    AdjustWindowRectEx(&ncrc,
-                       style,
-                       false, // Add a field to WindowSpec to add native menu bars
-                       exStyle);
+
+    auto& winApi = system()->winApi();
+    if (winApi.GetWindowDpiAwarenessContext) {
+      UINT dpi = winApi.GetDpiForWindow(self->m_hwnd);
+      winApi.AdjustWindowRectExForDpi(&ncrc,
+                              style,
+                              false, // Add a field to WindowSpec to add native menu bars
+                              exStyle,
+                              dpi);
+    } else {
+      AdjustWindowRectEx(&ncrc,
+                        style,
+                        false, // Add a field to WindowSpec to add native menu bars
+                        exStyle);
+    }
 
     if (rc.x != CW_USEDEFAULT)
       rc.x += ncrc.left;
@@ -2541,6 +2560,14 @@ LRESULT CALLBACK WindowWin::staticWndProc(HWND hwnd, UINT msg, WPARAM wparam, LP
 
     if (wnd && wnd->m_hwnd == nullptr)
       wnd->m_hwnd = hwnd;
+      
+    // Enable scaling for titlebar with legacy per-monitor dpi awareness steeting
+    auto& winApi = system()->winApi();
+    if (winApi.GetWindowDpiAwarenessContext) {
+      DPI_AWARENESS_CONTEXT dpiContext = winApi.GetWindowDpiAwarenessContext(hwnd);
+      if (!winApi.AreDpiAwarenessContextsEqual(dpiContext, DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2))
+        winApi.EnableNonClientDpiScaling(hwnd);
+    }
   }
   else {
     wnd = reinterpret_cast<WindowWin*>(GetWindowLongPtr(hwnd, GWLP_USERDATA));
--- a/os/win/window.cpp
+++ b/os/win/window.cpp
@@ -266,7 +266,7 @@ WindowWin::WindowWin(const WindowSpec& spec)
 
   if (winApi.GetWindowDpiAwarenessContext) {
     double scaleFactor = winApi.GetDpiForWindow(m_hwnd) / (double) USER_DEFAULT_SCREEN_DPI;
-    m_scale = std::max(1, (int)(m_baseScale * scaleFactor + 0.51));
+    m_scale = std::ceil(m_baseScale * scaleFactor);
   }
 
   // This flag is used to avoid calling T::resizeImpl() when we
--- a/os/win/window.cpp
+++ b/os/win/window.cpp
@@ -977,7 +977,7 @@ LRESULT WindowWin::wndProc(UINT msg, WPARAM wparam, LPARAM lparam)
 
     case WM_DPICHANGED: {
       double scaleFactor = HIWORD(wparam) / (double) USER_DEFAULT_SCREEN_DPI;
-      m_scale = std::max(1, (int)(m_baseScale * scaleFactor + 0.51));
+      m_scale = std::ceil(m_baseScale * scaleFactor);
 
       RECT* prcNewWindow = (RECT*)lparam;
       SetWindowPos(m_hwnd,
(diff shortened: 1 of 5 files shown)


dacap:
Thanks a lot @lampysprites for the dedication to see this issue. I'll try to review this soon, not for the next release but probably for the other, and I'll try to merge this on beta branch (so we can release a preview of the fix to a small number of users first).

martincapello:
I've been testing these changes with the hello_laf extension and found an issue when having 2 displays.
Scenario:
Display 1: 2560 x 1440 at 175% scale
Display 2: 1920 x 1080 at 150% scale 

Reproduction steps:
1. Run the hello_laf exe in display 1.
2. Press `2` to set the window scale to 2.
3. Move the windows to display 2.
4. Press `1`.

The size of the pixels in the client area of the window are smaller than it should.

The expected result should be the same as when doing this:
1. Run the hello_laf exe in display 1.
2. Move the window to display 2.

I think that this can be fixed by replacing this:
```c++
void WindowWin::setScale(int scale)
{
  double scaleFactor = m_scale / (double) m_baseScale;
  m_baseScale = scale;
  m_scale = scale * scaleFactor;
```
by 
```c++
void WindowWin::setScale(int scale)
{
  m_baseScale = scale;
  auto& winApi = system()->winApi();
  if (winApi.GetWindowDpiAwarenessContext)
    m_scale = MulDiv(m_baseScale, winApi.GetDpiForWindow(m_hwnd), USER_DEFAULT_SCREEN_DPI);
  else
    m_scale = scale;
```

os/win/window.cpp
```diff
@@ -263,6 +264,11 @@ WindowWin::WindowWin(const WindowSpec& spec)
 
   SetWindowLongPtr(m_hwnd, GWLP_USERDATA, reinterpret_cast<LONG_PTR>(this));
 
+  if (winApi.GetWindowDpiAwarenessContext) {
+    double scaleFactor = winApi.GetDpiForWindow(m_hwnd) / (double) USER_DEFAULT_SCREEN_DPI;
+    m_scale = std::max(1, (int)(m_baseScale * scaleFactor + 0.51));
```
martincapello, line 269: I think we could use the ceil funciton instead:
```suggestion
    m_scale = std::ceil(m_baseScale * scaleFactor);
```

os/win/window.cpp
```diff
@@ -361,7 +367,10 @@ os::ColorSpaceRef WindowWin::colorSpace() const
 
 void WindowWin::setScale(int scale)
 {
-  m_scale = scale;
+  double scaleFactor = m_scale / (double) m_baseScale;
+  m_baseScale = scale;
+  // round to nearest integer, rounding 0.5 up
+  m_scale = std::max(1, (int)(scale * scaleFactor + 0.51));
```
martincapello, line 373: Same here:
```suggestion
  m_scale = std::ceil(m_baseScale * scaleFactor);
```

os/win/window.cpp
```diff
@@ -966,6 +975,21 @@ LRESULT WindowWin::wndProc(UINT msg, WPARAM wparam, LPARAM lparam)
       }
       break;
 
+    case WM_DPICHANGED: {
+      double scaleFactor = HIWORD(wparam) / (double) USER_DEFAULT_SCREEN_DPI;
+      m_scale = std::max(1, (int)(m_baseScale * scaleFactor + 0.51));
```
martincapello, line 980: And here:
```suggestion
      m_scale = std::ceil(m_baseScale * scaleFactor);
```

martincapello (commented):
This works great! I've just suggested changing the use of `std::max` by `std::ceil` in three places, because I think it is a nice use case for it.

martincapello (commented):
While reviewing the [Aseprite changes](https://github.com/aseprite/aseprite/pull/4934) that make use of this changes I've found this issue:  https://github.com/aseprite/aseprite/pull/4934#pullrequestreview-2722938810

Which is something that should be fixed in laf. The issue can be reproduced modifying the hello_laf example a bit, by just setting the window as floating. Then while this modified version of the hello_laf example is running, try to change the display scale from Windows configuration and you will see that the window is not updated until it is moved/resized.

Will try to create a bare minimum WinAPI app similar to hello_laf to see what happens in this case.

martincapello:
This is basically a default Windows Desktop Application generated by Microsoft Visual Studio Community 2022 with some modifications:
- Removal of some unneeded things (like the about dialog, and some code that is not needed to prove the issue). 
- It sets the SetProcessDpiAwarenessContext to DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2.
- Replace the use of CreateWindowW by CreateWindowEx, to set the WS_EX_TOOLWINDOW style of the window.

[dpiaware_issue.zip](https://github.com/user-attachments/files/19510560/dpiaware_issue.zip)

This solution demonstrates that the WM_DPICHANGED message is not arriving when the user changes the display's DPI and the window has the WS_EX_TOOLWINDOW style set. If this style is not set, it works as expected.

Will try to send this to Microsoft and see what happens...

lampysprites:
Thanks a lot for spending time on this! I'm really sorry it turned into this kind of troubleshooting🥲
I've applied your changes with std::ceil and updated laf hash in the aseprite pr with that, is there something else?

martincapello:
> Thanks a lot for spending time on this! I'm really sorry it turned into this kind of troubleshooting🥲 

No worries! Thanks to you actually for bringing this new feature.

> I've applied your changes with std::ceil and updated laf hash in the aseprite pr with that, is there something else?

Nothing else by me, maybe after David's review he could point out some changes.

dacap:
I'm going to try to merge this (probably rebasing and pre-commiting to beta) after some testing.

dacap:
Thanks @lampysprites for the patience and your work on this issue.

I was testing this PR with a little patch just for debugging purposes here:
```diff
diff --git a/os/win/window.cpp b/os/win/window.cpp
index 6ee4a33..583581d 100644
--- a/os/win/window.cpp
+++ b/os/win/window.cpp
@@ -1023,6 +1023,8 @@ LRESULT WindowWin::wndProc(UINT msg, WPARAM wparam, LPARAM lparam)
       double scaleFactor = HIWORD(wparam) / (double)USER_DEFAULT_SCREEN_DPI;
       m_scale = std::ceil(m_baseScale * scaleFactor);
 
+      PRINTARGS("WM_DPICHANGED m_scale=", m_baseScale, "*", scaleFactor, "=", m_scale, "(", title(), ")");
+
       RECT* prcNewWindow = (RECT*)lparam;
       SetWindowPos(m_hwnd,
                    nullptr,
```

When we move each window between monitors the signal is received for both kind of windows (main window and the floating ones):

https://github.com/user-attachments/assets/e67a7e02-9984-4205-8f38-4ef8e4eae034

But I've found the problem that @martincapello commented:

https://github.com/user-attachments/assets/d1f51b79-bef7-4e9c-b7ef-3f7c20ce7fa3

And as Aseprite uses floating windows for children windows when the "multiple windows" interface is enabled
image
it is a problem as these don't receive the DPI change signal.

Another problem I've found is in the `custom_window` example:

https://github.com/user-attachments/assets/aac36b32-17ce-41b6-81e8-c8575ff54cec

It looks like hit areas are not calculated correctly, but I've tested Aseprite custom windows and the hit areas are correctly calculated, so it might be something to fix only in `custom_window` specifically.

I'll try to cherry-pick this PR squashed running a pre-commit (prek) hook before in the beta branch. The API looks nice, I'm not sure if it should be `scale/baseScale()` or `scale/realScale()` (so `scale()` is the user-defined scale). I'll see if I rename those methods later in further commits.

dacap:
Merged in https://github.com/aseprite/laf/commit/b1e5a4f6826c5ab7951315cbb2123f06e3697535, thanks again @lampysprites 🙏

PR #128, opened 2025-01-17
Add null UUID definition to use for comparing

martincapello:
This PR just defines a "null" UUID  constant by introducing a static const member to the Uuid class.

### DIFF
--- a/base/CMakeLists.txt
+++ b/base/CMakeLists.txt
@@ -56,6 +56,7 @@ set(BASE_SOURCES
   thread.cpp
   thread_pool.cpp
   time.cpp
+  uuid.cpp
   version.cpp)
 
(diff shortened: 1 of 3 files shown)


martincapello:
This is not needed anymore actually...so feel free to do whatever you want with this.

dacap:
I'll close this as it's not required and the expected behavior is that just creating a new `Uuid()` with the default contructor will create an empty `Uuid()`:

https://github.com/aseprite/laf/blob/412139cee7283ffebcf701ed47111a4eb016d239/base/uuid_tests.cpp#L13-L19

Generally `uuid == Uuid()` should be enough to check if an UUID is empty, but we could add a new member function `bool empty() const` if needed.

PR #129, opened 2025-01-30
[riscv64] Add Arch

Xeonacid:
Enable building on riscv64.


dacap:
Thanks @Xeonacid!

PR #130, opened 2025-02-04
Create a new laf-dlgs-proc executable to run the file dialog in a child process

dacap:
We want to fix https://github.com/aseprite/aseprite/issues/4412 in some way, this is not yet ready because it looks like a child process cannot use `IModalWindow::Show` with a parent process HWND handle.


martincapello (commented):
As mentioned in https://github.com/aseprite/aseprite/pull/5450#pullrequestreview-3311209369 seems that the dlg_process is not aware about the parent process status. So killing the parent process is not killing the dlg_process.

PR #131, opened 2025-02-06
Fix NSApplication initialization to prevent some issues on macOS

martincapello:
By calling [NSApplication run] and then [NSApplication stop] we are able to initialize the application instance and prevent issues in fullscreen mode (fix aseprite/aseprite#4795)

### DIFF
--- a/os/osx/app.mm
+++ b/os/osx/app.mm
@@ -52,7 +52,14 @@ public:
 
   void finishLaunching()
   {
-    [m_app finishLaunching];
+    // [m_app run] can only be called once.
+    if (![[NSRunningApplication currentApplication] isFinishedLaunching])
+      // Note that the [m_app run] call doesn't block because we are calling
+      // [NSApp stop] from [AppDelegateOSX applicationDidFinishLaunching]. We only
+      // need the application's initialization done inside run to prevent issues
+      // such as: https://github.com/aseprite/aseprite/issues/4795
+      [m_app run];
+
     [m_appDelegate resetCliFiles];
   }
 
--- a/os/osx/app.mm
+++ b/os/osx/app.mm
@@ -52,13 +52,24 @@ public:
 
   void finishLaunching()
   {
-    // [m_app run] can only be called once.
-    if (![[NSRunningApplication currentApplication] isFinishedLaunching])
-      // Note that the [m_app run] call doesn't block because we are calling
-      // [NSApp stop] from [AppDelegateOSX applicationDidFinishLaunching]. We only
-      // need the application's initialization done inside run to prevent issues
-      // such as: https://github.com/aseprite/aseprite/issues/4795
-      [m_app run];
+    id runningApp = [NSRunningApplication currentApplication];
+    // [m_app run] must be called once, if the app didn't finish launching yet.
+    if (![runningApp isFinishedLaunching]) {
+      // The run method must be called in GUI mode only, otherwise the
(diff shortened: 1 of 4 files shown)


os/osx/app.mm
```diff
@@ -52,7 +52,14 @@ void setAppMode(AppMode appMode)
...
-    [m_app finishLaunching];
+    // [m_app run] can only be called once.
+    if (![[NSRunningApplication currentApplication] isFinishedLaunching])
+      // Note that the [m_app run] call doesn't block because we are calling
+      // [NSApp stop] from [AppDelegateOSX applicationDidFinishLaunching]. We only
+      // need the application's initialization done inside run to prevent issues
+      // such as: https://github.com/aseprite/aseprite/issues/4795
+      [m_app run];
+
     [m_appDelegate resetCliFiles];
```
martincapello, line 63: Should we put this line under the scope of the if clause as well?

os/osx/app.mm
```diff
@@ -52,7 +52,14 @@ void setAppMode(AppMode appMode)
...
   void finishLaunching()
   {
-    [m_app finishLaunching];
+    // [m_app run] can only be called once.
+    if (![[NSRunningApplication currentApplication] isFinishedLaunching])
+      // Note that the [m_app run] call doesn't block because we are calling
+      // [NSApp stop] from [AppDelegateOSX applicationDidFinishLaunching]. We only
+      // need the application's initialization done inside run to prevent issues
+      // such as: https://github.com/aseprite/aseprite/issues/4795
+      [m_app run];
```
dacap, line 61: Just add {} to contain the whole comment+statement:
```suggestion
    if (![[NSRunningApplication currentApplication] isFinishedLaunching]) {
      // Note that the [m_app run] call doesn't block because we are calling
      // [NSApp stop] from [AppDelegateOSX applicationDidFinishLaunching]. We only
      // need the application's initialization done inside run to prevent issues
      // such as: https://github.com/aseprite/aseprite/issues/4795
      [m_app run];
    }
```

os/osx/app.mm
dacap: Just in case, we should test if dropping files to the app icon in the dock works. Probably it should be tested with the final Aseprite.app bundle, not only the `bin/aseprite` binary. I could give a try next week.

dacap (changes requested):

dacap:
Please @martincapello follow this rule https://github.com/aseprite/aseprite/blob/bbe656c856453b6a3d5cd617f8312417bcf11035/docs/CODING_STYLE.md?plain=1#L54-L59 even if those multiple lines are just comments.

martincapello:
> Please @martincapello follow this rule https://github.com/aseprite/aseprite/blob/bbe656c856453b6a3d5cd617f8312417bcf11035/docs/CODING_STYLE.md?plain=1#L54-L59 even if those multiple lines are just comments.

I saw your comment just after I pushed the last commit...

os/osx/app.mm
martincapello: I've tested with the binary with and without this fix, and it doesn't work either way. So it seems that we can only test this with the final bundle?

martincapello: I've just kind of tested this by vandalizing a v1.3.12 bundle (replaced the release binary with a compiled one 😬) and it worked...not sure if it will work by creating the bundle, but seems like it.

PR #133, opened 2025-03-19
Add 'is_absolute_path()' function

Gasparoken:
Part of the solution of issue aseprite/aseprite#5066

### DIFF
--- a/base/fs.cpp
+++ b/base/fs.cpp
@@ -1,5 +1,5 @@
 // LAF Base Library
-// Copyright (c) 2021-2024 Igara Studio S.A.
+// Copyright (c) 2021-2025 Igara Studio S.A.
 // Copyright (c) 2001-2018 David Capello
 //
 // This file is released under the terms of the MIT license.
@@ -34,6 +34,18 @@ const std::string::value_type* path_separators = "\\/";
 #else
 const std::string::value_type* path_separators = "/";
 #endif
+
+bool is_absolute_path(const std::string& path)
+{
+#if LAF_WINDOWS
+  return (path.size() > 2 && std::isalpha(path[0]) && path[1] == ':' &&
+          path[2] == path_separator) ||
+         // UNC network path
+         (path.size() > 1 && path[0] == path_separator && path[1] == path_separator);
+#else
+  return !path.empty() && path[0] == path_separator;
+#endif
+}
 
 void make_all_directories(const std::string& path)
 {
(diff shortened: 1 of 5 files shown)


dacap:
Please check that this PR now contains a commit with changes I've made in bd36a113081aa95da8d03b14a7627083e936c06b, and I'm the author (?)

base/fs.cpp
```diff
@@ -194,7 +205,8 @@ std::string get_relative_path(const std::string& filename, const std::string& ba
   auto itFrom = baseDirs.begin();
   auto itTo = toParts.begin();
 
-  while (itFrom != baseDirs.end() && itTo != toParts.end() && *itFrom == *itTo) {
+  while (itFrom != baseDirs.end() && itTo != toParts.end() &&
+         base::string_to_lower(*itFrom) == base::string_to_lower((*itTo))) {
```
dacap, line 209: We cannot do this on Linux (and sometimes depending on the configuration of the macOS file system), but I'd be worried only for Linux which the file system is case sensitive.

dacap (changes requested):

base/fs.cpp
```diff
@@ -35,6 +39,46 @@ const std::string::value_type* path_separators = "\\/";
...
+#if LAF_MACOS
+static bool* g_isfilesystemCaseSensitive = nullptr;
+
+bool isFilesystemCaseSensitive() {
+  if (g_isfilesystemCaseSensitive)
+    return *g_isfilesystemCaseSensitive;
+  std::string path1 = get_current_path() + "/filesystemTest.txt";
+  std::string path2 = get_current_path() + "/FilesystemTest.TXT";
+
+  std::ofstream file1(path1);
```
dacap, line 51: This can fail in a lot of different ways (e.g. we don't have access to write in the current path). I'd try to avoid checking this on macOS, or investigate the correct way to check if the file system is case sensitive (it might vary from file system/mount point, so I'm not sure if I'd worry about this).

Gasparoken: Ok. In conclusion, should we keep the macOS case-sensitive format as it was originally? Or adopt the new case-insensitive file name comparison, as described in the first pull request?

dacap: I think something like https://github.com/aseprite/laf/commit/3651931d55c4b9380c3a2004900e97829c5c611a is enough, Linux being the only platform where case-sensitive is allowed, and Windows/macOS case-insensitive.

PR #136, opened 2025-04-01
[win] Use SW_SHOWMAXIMIZED when first showing a maximized window (fix #5018)

ckaiser:
Fixes [#5018](https://github.com/aseprite/aseprite/issues/5018), I'm not 100% sold on the exact flag method or variable names, but the fix is sound.

Here's a video showing the change (recorded at 120 fps, then slowed down 5x)

Before:

https://github.com/user-attachments/assets/0cd4f1c1-be10-4710-aea3-03d475f9b158

---

After: 

https://github.com/user-attachments/assets/bee59769-a0f5-4607-b8c4-7493518d2895

### DIFF
--- a/os/win/window.cpp
+++ b/os/win/window.cpp
@@ -397,8 +397,16 @@ bool WindowWin::isVisible() const
 void WindowWin::setVisible(bool visible)
 {
   if (visible) {
-    if (m_activate)
-      ShowWindow(m_hwnd, SW_SHOWNORMAL);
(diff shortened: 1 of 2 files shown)


dacap:
It looks great, I tried on debug mode and didn't notice a difference, but then I was able to record the difference in release mode, I'm not sure if there is another issue in the middle, but it looks like this fixes this problem 👍

PR #138, opened 2025-04-06
Add configurable font hinting API

hwabis:

I was looking at https://github.com/aseprite/aseprite/issues/4931, and found the issue could be fixed when this adding after the following line:
https://github.com/aseprite/laf/blob/cc62448d580bd415c9e7259b99ac1986ed7c4212/text/skia_text_blob_shaper.cpp#L125
`skFont.setHinting(SkFontHinting::kNone);`

However, I assume we do not want hinting to be disabled by default / all the time. So this MR is setting up the necessary API for hinting to be configurable through the Aesprite app (I imagine through the top bar when the Text Tool is used).

### DIFF
--- a/text/font.h
+++ b/text/font.h
@@ -12,6 +12,7 @@
 #include "base/ints.h"
 #include "base/ref.h"
 #include "gfx/fwd.h"
+#include "text/font_hinting.h"
 #include "text/font_type.h"
 #include "text/fwd.h"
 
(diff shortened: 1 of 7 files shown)


dacap:
I'll merge this and add the missing changes to make EmptyFont compile.

dacap:
I'll redesign the font selector to be able to choose this hinting property.

PR #139, opened 2025-04-11
[osx] Wake up queue after drag & drop operations

martincapello:
This PR adds some code to send an empty event for waking up the events queue in case it was waiting for events. This is useful for flushing pending application events (like painting events) that might have been generated during any of the drag and drop operations.
This changes are needed to make a part of aseprite/aseprite#5106 work (the item that says: "Avoid creating/enqueing os::Event from the ui/app layers to the laf-os").


martincapello:
Closing, no need for this anymore.

PR #140, opened 2025-04-16
Add method to get the native color channels order

martincapello:
Introduced to fix the "Avoid creating a temporal surface in static WEBP_CSP_MODE colorMode()..." item from https://github.com/aseprite/aseprite/issues/4777

### DIFF
--- a/os/skia/skia_surface.cpp
+++ b/os/skia/skia_surface.cpp
@@ -62,6 +62,13 @@ static SkCanvas::SrcRectConstraint to_constraint(const Paint* paint)
   return SkCanvas::kStrict_SrcRectConstraint;
 }
 
+// static
+Surface::ColorChannelsOrder Surface::getNativeColorChannelsOrder()
+{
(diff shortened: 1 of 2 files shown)


os/skia/skia_surface.cpp
```diff
@@ -62,6 +62,13 @@ static SkCanvas::SrcRectConstraint to_constraint(const Paint* paint)
   return SkCanvas::kStrict_SrcRectConstraint;
 }
 
+// static
+Surface::ColorChannelsOrder Surface::getNativeColorChannelsOrder()
+{
+  return kN32_SkColorType == kRGBA_8888_SkColorType ? ColorChannelsOrder::RGB :
+                                                      ColorChannelsOrder::BGR;
```
dacap, line 69: I think these values (`kN32_SkColorType` and `kRGBA_8888_SkColorType`) probably something that can be solved with a `#if` or `if constexpr`?

dacap: Dont worry, I’ll try something later to have a static check that these constants have the two possible expected values.

PR #141, opened 2025-04-22
[osx] Fix scaling issues when painting dirty areas

martincapello:
The SkiaWindowOSX::paintGC was using integer units to calculate the scaled size portion of the source image that had to be updated. This caused that for odd rect sizes when using an even scale the rendering was offset by one pixel (one sub-pixel actually when scale was 2 for instance).

For reference: this issue was reported in a comment from another issue: https://github.com/aseprite/aseprite/issues/5095#issuecomment-2815292686

Steps to reproduce the issue:
1) Having the following settings:
     Screen scaling: 200%
     UI elements scaling: 100%
2) Snap the main window to the top of the screen (the right corner should do it as well).
3) Unsnap the window by dragging it from its title bar (this is the "interesting" part...unsnapping the window will make it have an odd size, not the size it had before snapping).
4) Move the mouse cursor around the bottom of the editor and the top of the timeline. You should see how the rendering of some lines of the window is offset by 1 pixel (1 subpixel actually because we are using 200% screen scaling).

### DIFF
--- a/os/skia/skia_window_osx.h
+++ b/os/skia/skia_window_osx.h
@@ -45,7 +45,7 @@ public:
   void onChangeBackingProperties() override;
 
 private:
-  void paintGC(const gfx::Rect& rect);
+  void paintGC(const gfx::RectF& rect);
 
   bool m_closing = false;
 
--- a/os/skia/skia_window_osx.mm
+++ b/os/skia/skia_window_osx.mm
@@ -192,7 +192,7 @@ void SkiaWindowOSX::onChangeBackingProperties()
     setColorSpace(colorSpace());
 }
 
-void SkiaWindowOSX::paintGC(const gfx::Rect& rect)
+void SkiaWindowOSX::paintGC(const gfx::RectF& rect)
 {
   if (!this->isInitialized())
     return;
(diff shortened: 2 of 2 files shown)


dacap:
It works 👍 it looks like a new bug on macOS with this new feature to dock windows on each side (should we report to Apple?). I'll merge this fix, probably we should investigate the possibility to round the window size to a multiple of `m_scale` in a future, just like in `WindowOSXObjc initWithImpl:spec:` (but it's not urgent):

https://github.com/aseprite/laf/blob/18a7d47675e865376700a65010698c93344d1db3/os/osx/window.mm#L73-L77

martincapello:
> It works 👍 it looks like a new bug on macOS with this new feature to dock windows on each side (should we report to Apple?).

Yeah, it seems like a macOS bug to me too...maybe we should report the bug, will see if I can make a simple demo app to reproduce it (no laf, just using appkit and objective-c).

> I'll merge this fix, probably we should investigate the possibility to round the window size to a multiple of `m_scale` in a future, just like in `WindowOSXObjc initWithImpl:spec:` (but it's not urgent):

Actually I had tried another fix that was like that and worked as well. I just went for this one because I realized that the `SkRect::MakeXYWH` parameters were of float type. But yeah, I'm fine with using any approach that works.

martincapello:
> Probably Apple will not consider this a macOS bug...

Well, then maybe it doesn't worth it to report it...so I won't do it, unless you say so.

PR #145, opened 2025-05-13
[x11] Reduce the MotionNotify events processing rate

### LINKED ISSUE #5028: Lag on Linux
I have noticed general app lag when doing pretty much anything in Aseprite, pretty sure the lag wasn't present in earlier versions.
Dragging windows is sluggish and sometimes lags behind, but what concerns me the most is that when drawing something on a canvas and frequently using Undo, it becomes really sluggish and slow, after that pretty much everything lags behind for a while and it's really annoying.

Haven't been able to test it on a different OS than linux, but if anyone has or had similar issues, would be nice to share them here.

Edit: after working in a 128x128 canvas for a tilemap, simply dragging around the canvas becomes super laggy as well and it's annoying.

martincapello:
By reducing the processing rate of X11's MotionNotify events we remove a lot of pressure from the application, because it doesn't have to keep up with all the MotionNotify events generated any more.

Should fix https://github.com/aseprite/aseprite/issues/5028. I say "should" because I was able to reproduce it just for the preferences dialog, so I'm not sure if it will fix the other cases reported by the users.

### DIFF
--- a/os/x11/event_queue.cpp
+++ b/os/x11/event_queue.cpp
@@ -132,6 +132,8 @@ void EventQueueX11::getEvent(Event& ev, double timeout)
   // key pressed.
   const bool removeRepeats = (!WindowX11::translateDeadKeys());
 
+  XEvent lastXMotion;
+  lastXMotion.type = 0;
   for (int i = 0; i < events; ++i) {
     XNextEvent(display, &event);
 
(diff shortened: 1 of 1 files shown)


dacap:
The only problem I find with this PR is that we are discarding a lot of mouse/stylus information to paint on the canvas. It might be a solution if we can enable this behavior when we are resizing something (a window, a selection, a slice, etc.) but not when we need all the motion events.

martincapello:
Got the point. The only thing that makes me worry a bit is that Aseprite performance depends on the MotionNotify events rate. If the user has a fast enough hardware, Aseprite might not be able to process all the events on time to make it look smoother. So at the end, the hardware that generates the appropriate rate of events would be the better. 
Some comments about this from a user: https://github.com/aseprite/aseprite/issues/5028#issuecomment-2908179205

dacap:
I'm closing this PR because the resize issue was fixed in https://github.com/aseprite/laf/commit/35d0bbd919f7d5a2f0b296b4670a77adfe4eb3e8 processing `ConfigureNotify` in a different way: when one `ConfigureNotify` message is received we consume all queued `ConfigureNotify` messages for the same window, and finally resize the window just one time.

To fix any other lag issue related to painting in the canvas I'd focus more in the rendering side and improving the painting logic in a background thread if possible (to avoid losing mouse/stylus events in the middle). But that is already for the Aseprite repo.

PR #147, opened 2025-06-12
[win] Add IMM support

Cerallin:
* Call TranslateMessage() when WM_KEYDOWN and it is not a dead key
* Set IMM form position: below cursor
* Add CMake option: LAF_WITH_IMM, default off

The code about TranslateMessage() still needs discussion. Am I doing right?
And I would like to know if there is a good way to get the caret position, which is used for IME positioning.

Ref: https://github.com/aseprite/aseprite/issues/1767


dacap:
The `os::System::setTranslateDeadKeys()` function is used by Aseprite to indicate when we are inside a input field. For example, if we are outside a input field, `setTranslateDeadKeys(false)` is called, so dead keys are not received, we need only keyboard shortcuts as they are. When we focus a text input field, then `setTranslateDeadKeys(true)` is called to start receiving dead keys/special unicode chars.

There is a comment in this member function:

https://github.com/aseprite/laf/blob/e04e7c52e3b82134f714b2472cf526af5f186b53/os/system.h#L196-L202

Probably we could just refactor `setTranslateDeadKeys()` to something like:
```c++
class System {
...
  virtual void setTextInput(bool state, const gfx::Point& screenCaretPos = {}) = 0;
...
};
```

Cerallin:
@dacap It seems much easier than I thought to check if user is inputting with IME. Please have a look: 
https://learn.microsoft.com/en-us/windows/win32/learnwin32/keyboard-input
https://learn.microsoft.com/en-us/windows/win32/api/imm/nf-imm-immgetvirtualkey

The key point is to handle messages with virtual key value of `VK_PROCESSKEY`. Should I still work on exposing an API to indicate if we are inside an input field or not?

dacap:
> Should I still work on exposing an API to indicate if we are inside an input field or not?

I think the main thing is that the app (not laf) is the only one that can tell us if we are inside a text field or not, which means what we should do with this `VK_PROCESSKEY`. For example, if we are not in a text field, we should ignore `VK_PROCESSKEY`.

Cerallin:
@dacap I’ve finished making the requested changes. Please take a look whenever it’s convenient.

dacap:
Thanks a lot for this PR @Cerallin, so far it's a good starting point to integrate in laf. I'll merge as it is and make some minor adjustments and more testing in a near future. I think we could start compiling some beta version of Aseprite with `LAF_WITH_IME=ON` in a near future.

PR #151, opened 2025-06-23
Recover Windows Vista compatibility

martincapello:
Fix aseprite/aseprite#5211


dacap:
Is this required for libarchive only? Because libarchive is in aseprite repo (not in laf repo).

PR #152, opened 2025-06-24
fix: check $ID alongside $ID_LIKE in misc/platform.sh

johnbotris:
in archlinux, `etc/os-release` does not define `ID_LIKE`, only `ID`. the build script fails as a result of only checking `ID_LIKE`, because `$cpu` ends up being unbound, leading to unfound skia binaries.

this fix is for aseprite/aseprite#5246


dacap:
Thanks @johnbotris, I'll merge this :+1:

PR #154, opened 2025-09-03
Fix newlines causing the textHeight and baseline to use fallback font

ckaiser:
Fixes aseprite/aseprite#5408

Tooltips in aseprite are one of the few things that have `\n` newline characters, so because of the changes introduced in [this commit](https://github.com/aseprite/aseprite/commit/af6e8b65c3c8a3fddba6bc95c594265c2fd6256c) the text inside those was using the fallback text height (which is by default now like twice the size of the regular one, which made it so noticeable).

The fix is now we avoid doing the processing for the most common line break codepoints and we only change the `textHeight` and `baseline` values to the fallback ones when the font actually finds a glyph for the codepoint.

### DIFF
--- a/text/sprite_text_blob_shaper.cpp
+++ b/text/sprite_text_blob_shaper.cpp
@@ -97,6 +97,10 @@ TextBlobRef SpriteTextBlob::MakeWithShaper(const FontMgrRef& fontMgr,
       if (chr == 0)
         break;
 
+      // Do not process newlines/control characters
+      if (chr >= 10 && chr <= 20)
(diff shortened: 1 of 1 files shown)


ckaiser:
I think trying to make the code more robust I ended up making it more brittle, since we can't get emoji glyphs from the fallback font (I think Skia falls back even further, or there's a separate emoji font) it always comes back as 0, I think skipping the new line/control characters is gonna be enough for at least this fix without over-complicating things or causing other side effects.

PR #155, opened 2025-09-03
Fix color space handling in Skia surfaces

Gasparoken:


dacap:
Just tested and it fixes the color bug in the canvas/sprite editor 👍 LGTM

PR #161, opened 2025-10-31
Fix bounds and workarea on additional screen in macOS

Gasparoken:


dacap:
Works great, just tested listscreens and multiple_windows examples 👍

PR #163, opened 2025-11-07
Allow drawing the selected text background earlier

martincapello:
This PR modifies the order in which some things are done when each run of text is processed when the shaping function is executed:
- Now, the selected text background drawing is delegated before drawing any character. In this way if the glyphs in a run overlap with each other the background drawing of the current glyph won't overwrite the previous glyph.
- Now, if there is a background color set, it is painted before drawing any character and before delegating the selected text background painting.

Both of the above things are possible because now we first calculate the union of all glyphs bounds.

### DIFF
--- a/text/draw_text_shaper.cpp
+++ b/text/draw_text_shaper.cpp
@@ -1,5 +1,5 @@
 // LAF Text Library
-// Copyright (c) 2024  Igara Studio S.A.
+// Copyright (c) 2024-2025  Igara Studio S.A.
 //
 // This file is released under the terms of the MIT license.
 // Read LICENSE.txt for more information.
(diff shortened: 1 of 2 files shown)


text/draw_text_shaper.cpp
```diff
@@ -49,11 +50,31 @@ class AdapterBuilder : public TextBlob::RunHandler {
   void commitRunBuffer(TextBlob::RunInfo& info) override
   {
     if (info.clusters && info.glyphCount > 0) {
-      float advanceX = 0.0f;
+      auto selectionBounds = gfx::RectF();
+      auto textBounds = gfx::RectF();
```
dacap: Just use
```suggestion
      gfx::RectF selectionBounds;
      gfx::RectF textBounds;
```

text/draw_text_shaper.cpp
```diff
@@ -49,11 +50,31 @@ class AdapterBuilder : public TextBlob::RunHandler {
...
+      std::map<int, gfx::RectF> glyphsBounds;
```
dacap: Why a std::map and not a std::vector here? It looks like this variable is being used as a vector/array.

text/draw_text_shaper.cpp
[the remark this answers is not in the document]
martincapello: Oh, no, that was added automatically by vs code...I'll remove it, thank you!

text/draw_text_shaper.cpp
martincapello: You are right, I think I thought that I'd need random access to the elements...I'll change it. Thanks!

Actually that is possible with a vector too...so don't know what I was thinking...

text/draw_text_shaper.cpp
```diff
@@ -49,11 +49,31 @@ class AdapterBuilder : public TextBlob::RunHandler {
   void commitRunBuffer(TextBlob::RunInfo& info) override
   {
     if (info.clusters && info.glyphCount > 0) {
-      float advanceX = 0.0f;
+      gfx::RectF selectionBounds;
+      gfx::RectF textBounds;
+      std::vector<gfx::RectF> glyphsBounds;
+      for (int i = 0; i < info.glyphCount; ++i) {
+        auto bounds = info.getGlyphBounds(i);
+        glyphsBounds.push_back(bounds);
```
dacap: We can prealloc the expected size of `glyphsBounds`:
```suggestion
      std::vector<gfx::RectF> glyphsBounds(info.glyphCount);
      for (int i = 0; i < info.glyphCount; ++i) {
        auto bounds = info.getGlyphBounds(i);
        glyphsBounds[i] = bounds;
```

PR #165, opened 2025-11-10
[x11] Add movement threshold when detecting double clicks in X11

### LINKED ISSUE #112: x11 Double Click inconsistent because of zero movement threshold
LAF implements double click itself on x11 but it has no movement threshold, which causes dropping a lot of intended double clicks due to small mouse movements between the first and second click.

I noticed this when switching to linux and using aseprite after being on windows for awhile, using double click to select tile-wise was only working 50% of the time or less.

It should accumulate movement delta values until a threshold is reached before dropping the pending double click state. (a few pixels of screenspace should be a reasonable default I think Windows defaults to 4 horizontally and vertically counted separately)

Cammymoop:
Fixes #112
This makes double clicking in aseprite much less flakey on linux, for example when using the marque tool in tile mode, or navigating directories and opening files using the aseprite file dialogs. Without this I have to be very careful to keep the mouse still when double clicking and even then I usually have to try several times to get a double click to register.

Prevents the x11 implementation from resetting the click state on mouse movement until either x or y position has changed more than a certain radius from the last clicked position. Also checks the position again before triggering a double click action just in case the position has somehow changed in a way that wasn't captured and handled.

I set it up with a fixed radius of 8, since I'm not really sure how best to go about making it build configurable or actually dynamic. But this should be a huge improvement over the existing behavior with no threshold. For comparison with windows the default radius seems to be 2, but probably works differently for touch/tablet inputs, SDL defaults to 32 which seems a bit excessive so I picked something a bit more conservative but still hopefully useful if you happen to be using touch or tablet input.

### DIFF
--- a/os/x11/window.cpp
+++ b/os/x11/window.cpp
@@ -1199,14 +1199,19 @@ void WindowX11::processX11Event(XEvent& event)
         ev.setButton(button);
 
         if (event.type == ButtonPress) {
+          gfx::Point currentPos(event.xbutton.x / m_scale, event.xbutton.y / m_scale);
+
           if (m_doubleClickButton == button &&
-              base::current_tick() - m_doubleClickTick < LAF_X11_DOUBLE_CLICK_TIMEOUT) {
+              base::current_tick() - m_doubleClickTick < LAF_X11_DOUBLE_CLICK_TIMEOUT &&
+              std::abs(currentPos.x - m_doubleClickStartPos.x) < kDoubleClickThreshold &&
+              std::abs(currentPos.y - m_doubleClickStartPos.y) < kDoubleClickThreshold) {
             ev.setType(Event::MouseDoubleClick);
             m_doubleClickButton = Event::NoneButton;
           }
           else {
             m_doubleClickButton = button;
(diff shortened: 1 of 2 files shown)


Cammymoop:
Thanks for reviewing
> It looks like the pre-commit failed, [it's just the order of #include files](https://github.com/aseprite/laf/actions/runs/19243853568/job/55434343010#step:4:135), if you can amend your commit I'll merge this.

Fixed this

dacap:
Thanks @Cammymoop!

Cammymoop:
> Checking quickly the issue with wacom tablets on Linux it seems that in
> 
> ```
> void WindowX11::processX11Event(XEvent& event)
> {
>    ...
>    ...
> 
>     case ButtonRelease: {
>       // This can happen when the button press/release events are
>       // handled in XInput
>       if (event.xmotion.time == g_lastXInputEventTime)
>         break;   
> }
> ```
> 
> event.xmotion.time == g_lastXInputEventTime is always true, and the double click handling code is never executed.

If that's the case then it should break all click event handling, not just double click logic, shouldn't it? What behavior do you get using the tablet with Aseprite?

PR #174, opened 2026-05-04
Add the posibility to allow atomic move_file on Windows

Gasparoken:
Needed to fix issue aseprite/aseprite#5770 and needed to complete PR aseprite/aseprite#5782)

### DIFF
--- a/base/fs_win32.h
+++ b/base/fs_win32.h
@@ -1,5 +1,5 @@
 // LAF Base Library
-// Copyright (c) 2020-2025 Igara Studio S.A.
+// Copyright (c) 2020-present Igara Studio S.A.
 // Copyright (c) 2001-2018 David Capello
 //
 // This file is released under the terms of the MIT license.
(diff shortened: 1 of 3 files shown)


base/fs_win32.h
```diff
@@ -42,11 +42,38 @@ size_t file_size(const std::string& path)
   return (_wstat(from_utf8(path).c_str(), &sts) == 0) ? sts.st_size : 0;
 }
 
-void move_file(const std::string& src, const std::string& dst)
+void move_file(const std::string& src, const std::string& dst, bool overwrite)
 {
-  BOOL result = ::MoveFile(from_utf8(src).c_str(), from_utf8(dst).c_str());
-  if (result == 0)
-    throw Win32Exception("Error moving file");
+  if (overwrite) {
```
dacap: It looks like there is a [MoveFileEx](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-movefileexa) that can receive a MOVEFILE_REPLACE_EXISTING, would be that enough for the overwrite case?

There is an example here: https://learn.microsoft.com/en-us/windows/win32/fileio/creating-and-using-a-temporary-file where they also use MOVEFILE_COPY_ALLOWED which to copy/delete the file between different drives (e.g. if the temp path is in other drive):

> If the file is to be moved to a different volume, the function simulates the move by using the [CopyFile](https://learn.microsoft.com/en-us/windows/desktop/api/winbase/nf-winbase-copyfile) and [DeleteFile](https://learn.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-deletefilea) functions.

> If the file is successfully copied to a different volume and the original file is unable to be deleted, the function succeeds leaving the source file intact.

Gasparoken: Yes, that will be enough. Thanks for the advice. I also saw that there is: `Curl_rename` which uses `MoveFileEx` internally.

dacap:
@Gasparoken the pre-commit failed https://github.com/aseprite/laf/actions/runs/25382562718/job/74434598727#step:4:128

Gasparoken:
@dacap, done.

dacap:
@Gasparoken I have added some tests https://github.com/aseprite/laf/commit/fca7f614d7887e29c61eb2c2988bcb78b233dbae

PR #178, opened 2026-05-14
[win] Wake up event queue when pushing an event

ckaiser:
Fixes the issue by posting a message to the thread to wake it up.

CI failure on the skia-linked branches seems to be unrelated-ish? It's crashing because it's missing zlib.dll, it's the same reason that app_main_tests fails... is that why it's getting removed by the current github action? @dacap

### DIFF
--- a/os/win/event_queue.cpp
+++ b/os/win/event_queue.cpp
@@ -9,20 +9,23 @@
   #include "config.h"
 #endif
 
-#include <queue>
-
-#include <windows.h>
-
-#include "os/win/event_queue.h"
-
 #include "base/time.h"
+#include "os/win/event_queue.h"
 #include "os/win/ime_manager.h"
 
 namespace os {
 
+EventQueueWin::EventQueueWin() : m_sleeping(false)
+{
+  m_threadId = GetCurrentThreadId();
+}
+
 void EventQueueWin::queueEvent(const Event& ev)
 {
   m_events.push(ev);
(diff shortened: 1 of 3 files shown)


os/win/event_queue.cpp
```diff
@@ -9,20 +9,21 @@
...
 
+EventQueueWin::EventQueueWin()
+{
+  m_threadId = GetCurrentThreadId();
+}
+
 void EventQueueWin::queueEvent(const Event& ev)
 {
   m_events.push(ev);
+  PostThreadMessage(m_threadId, WM_NULL, 0, 0);
```
dacap: It'd be nice to call this function only when the event queue is blocked in `getEvent()`.

ckaiser: Yeah I wanted to make sure it was the right idea first. Only thing too is that for that we'd have to lock and unlock a mutex and I'm wondering if that's not slower than just shooting off a message, will test.

dacap: Probably just a bool (or std::atomic<bool>) would be necessary, I'm don't remember why the std::atomic was replaced with a mutex in the macOS case here https://github.com/aseprite/laf/issues/89 / https://github.com/aseprite/laf/pull/92

ckaiser: Should be good now, getting a little fancy with the atomics

PR #179, opened 2026-05-15
[x11] Fix file drops from Wayland sources

SanjoSolutions:
## Summary
- keep pending Xdnd drop state alive when XWayland sends XdndLeave immediately after XdndDrop
- retry file-data selection conversion with common Wayland/XDG portal targets when text/uri-list is unavailable
- tighten URI-list parsing for CRLF, comments, localhost file URIs, and plain-path fallback payloads

## Testing
- Built Aseprite with `cmake --build build --target aseprite`
- Installed locally and verified dragging a file from GNOME Files into an Aseprite window opens the file on Ubuntu 25.10 / GNOME 49 Wayland

Context: aseprite/aseprite#1410

### DIFF
--- a/os/x11/dnd.h
+++ b/os/x11/dnd.h
@@ -20,7 +20,12 @@ struct DndDataX11 {
   ::Window sourceWindow = 0;
   gfx::Point position;
   DropOperation supportedOperations = DropOperation::None;
+  DropOperation selectedOperation = DropOperation::None;
   std::vector<Atom> types;
+  std::vector<Atom> targetFallbacks;
+  Time dropTime = CurrentTime;
+  int targetFallbackIndex = 0;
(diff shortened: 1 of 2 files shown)


SanjoSolutions:
I have fixed the linter formatting errors.

dacap:
Sorry @SanjoSolutions, I saw that you've used Codex to create your patch, we're not accepting PRs co-authored by AI. I'll close this PR and add a note about this in the PR template (Aseprite already have the note).

SanjoSolutions:
Ok. I hope that it is still fixed, so other Linux users receive the same benefit. You can also copy the changes from the PR (I give you MIT-0 permissions).

While I don't care much about your choices, I would recommend to reconsider your policy and make it more reasonable.

Best regards
Jonas

PR #180, opened 2026-05-17
[win] Use OutputDebugString instead of stderr on Windows

ckaiser:
fputs isn't really visible in Visual Studio, this makes LOG calls actually show up when developing on Windows

Including windows.h first because it defines ERROR because of course it does, and then log.h undefs it :^]

### DIFF
--- a/base/log.cpp
+++ b/base/log.cpp
@@ -9,6 +9,10 @@
   #include "config.h"
 #endif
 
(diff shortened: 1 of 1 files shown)


dacap:
Some time ago I did something similar, but I cannot remember if it was in Aseprite or other project, I think I reverted the change because there was a double output in the debug console, but I don't see why this should be the case.

PR #181, opened 2026-05-18
Fix combo box lists do not expand in the Preference dialog on MacOS

Gasparoken:
This fixes a regression inserted in the commit https://github.com/aseprite/aseprite/commit/d9785e5c363a555fcda30ad0a256abc59f5f6550: PR: https://github.com/aseprite/aseprite/pull/5813

### DIFF
--- a/os/osx/window.mm
+++ b/os/osx/window.mm
@@ -1,5 +1,5 @@
 // LAF OS Library
-// Copyright (C) 2019-2025  Igara Studio S.A.
+// Copyright (C) 2019-present  Igara Studio S.A.
 // Copyright (C) 2012-2017  David Capello
 //
 // This file is released under the terms of the MIT license.
@@ -117,10 +117,6 @@
       [self center];
     }
 
-    if (spec->parent())
-      self.parentWindow =
-        (__bridge NSWindow*)static_cast<os::WindowOSX*>(spec->parent())->nativeHandle();
-
     [self makeKeyAndOrderFront:self];
 
     m_floating = spec->floating();
@@ -129,6 +125,17 @@
 
     if (spec->modal())
       self.level = NSModalPanelWindowLevel;
+
+    // If a window has a new child window (like a combo box list),
+    // it must be brought above its parent's z-order so it always
+    // appears on top, even when the parent is clicked/activated.
+    if (spec->parent()) {
+      NSWindow* parent =
+        (__bridge NSWindow*)static_cast<os::WindowOSX*>(spec->parent())->nativeHandle();
+      [self setParentWindow:parent];
+      if ([parent level] >= [self level])
+        [self setLevel:[parent level] + 1];
+    }
 
     // Hide the "View > Show Tab Bar" menu item
     if ([self respondsToSelector:@selector(setTabbingMode:)])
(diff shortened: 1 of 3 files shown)


martincapello (approved):
Works great and looks good to me!

Gasparoken:
I only corrected one comment after the revision.

os/osx/window.mm
```diff
@@ -130,6 +126,27 @@ - (WindowOSXObjc*)initWithImpl:(os::WindowOSX*)impl spec:(const os::WindowSpec*)
...
+        [(WindowOSXObjc*)parent childDidBecomeKey:self];
+      if (parent.level >= self.level)
+        self.level = parent.level + 1;
+    }
+
+    // Add Restore child window order when app is reactivated
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(appDidBecomeActive:)
+                                                 name:NSApplicationDidBecomeActiveNotification
+                                               object:nil];
```
dacap: Instead of using the `NSNotificationCenter`, probably we could use the `applicationDidBecomeActive` from our own `AppDelegate` to avoid adding a new observer for each window. And then we can re-order the windows there (or call a `WindowOSX` function to do the `appDidBecomeActive` job).

(Our `applicationDidBecomeActive` is already doing something for floating windows)

os/osx/window.mm
```diff
@@ -130,6 +126,27 @@ - (WindowOSXObjc*)initWithImpl:(os::WindowOSX*)impl spec:(const os::WindowSpec*)
     if (spec->modal())
       self.level = NSModalPanelWindowLevel;
 
+    m_savedChildOrder = [NSMutableArray array];
```
dacap: We could keep `m_savedChildOrder = nullptr` until the first child window is added.

os/osx/window.mm
dacap: Seeing our current `applicationWillResignActive`:
```objc++
- (void)applicationWillResignActive:(NSNotification*)notification
{
  for (id window : [NSApp windows]) {
    if ([window isKindOfClass:[WindowOSXObjc class]] && [window isFloating]) {
      [window setLevel:NSNormalWindowLevel];
      [window orderWindow:NSWindowAbove relativeTo:0];
    }
  }

  NSEvent* event = [NSApp currentEvent];
  if (event != nil)
    [ViewOSX updateKeyFlags:event];
}
```
Wouldn't be possible to reorder children just starting from `[NSApp windows]`? (without a our auxiliary `m_savedChildOrder` array?)

Gasparoken: > Instead of using the NSNotificationCenter, probably we could use the applicationDidBecomeActive from our own AppDelegate to avoid adding a new observer for each window. And then we can re-order the windows there (or call a WindowOSX function to do the appDidBecomeActive job).

Yes, you're right! I'll explore that option.

> Wouldn't be possible to reorder children just starting from [NSApp windows]? (without a our auxiliary m_savedChildOrder array?)

No, '[NSApp windows]' this array do not guarantee the correct windows order ([docs](https://developer.apple.com/documentation/appkit/nsapplication/windows)). On the other hand '[NSApp orderedWindows]' is an array of window objects arranged according to their front-to-back ordering on the screen, but even with `orderedWindows`, we need to remember the array when Aseprite loses focus, since I empirically verified that the z-order of child windows is lost. This makes me think it might be a Cocoa/Appkit bug, but I'm not sure.

PR #182, opened 2026-05-20
[win] Fix windows never receiving a mouseup event after moving/resizing

### LINKED ISSUE #5449: [Lua API] Resizing/moving dialogue window causes mouse button to be held down for canvas widget
**Describe the bug**
When moving or resizing a dialogue window, the `MouseEvent.button` from canvas widgets with an `onmousemove` function will continue to read as `MouseButton.LEFT` even if the button is no longer held down. This happens even with multiple canvases on different dialogue windows. 

This will fix itself if I click the mouse again, so its not a big issue. It only happens when when User Interface is set to multiple window mode. 

**To Reproduce**
1. Set User Interface to Multiple Window Mode.
2. Run the following test script:
```
dialog = Dialog {
    id = "dialog",
    title = "Dialog",
    autofit = -1,
    onclose = function()
        dialog = nil
    end
}

dialog:label{
    id = "label",
    text = "Mouse Button: NONE"
}

dialog:canvas{
    id = "canvas",
    autoscaling = true,
    width = 320,
    height = 320,
    onpaint=function(ev)
        local gc = ev.context
        gc.color = gc.theme.color.editor_face
        gc:fillRect(Rectangle(0,0,gc.width,gc.height))
        gc:drawThemeRect("editor_selected", 0,0,gc.width, gc.height)
    end,
    onmousemove=function(ev)
        local buttons = {"NONE","LEFT","RIGHT","MIDDLE","X1","X2"}
        dialog:modify{id="label", text = "Mouse Button: ".. buttons[ev.button+1]}
    end,
}

dialog:show{wait=false}
```
3. Resize or move the dialog window.
4. Release mouse button.
5. Move mouse cursor over the canvas area.
 
**Expected behavior**
Mouse Button should be read as `NONE` because the mouse button is no longer held down.

**Screenshots**
If applicable, add screenshots or a screen recording to help explain your problem.

Image

My mouse is no longer held down after I resize the window.

**Aseprite & System (please complete the following information):**
- Aseprite: v1.3.15.4
- System: Windows 11

**Additional context**
This only happens when User Interface is set to multiple window mode. Behavior is normal when in single window mode.

ckaiser:
Fixes [#aseprite/5449](https://github.com/aseprite/aseprite/issues/5449) and potentially a few other edge cases and event weirdness. Inspired by [the solution used by Qt](https://github.com/qt/qtbase/blob/239c54452fa60157c90901c8be8685048a65ad0a/src/plugins/platforms/windows/qwindowscontext.cpp#L1263).

### DIFF
--- a/os/win/window.cpp
+++ b/os/win/window.cpp
@@ -67,6 +67,18 @@
 #endif
 
 namespace os {
+namespace {
+void query_mouse_buttons(std::map<Event::MouseButton, bool>& map)
+{
+  std::vector<Event::MouseButton> buttons;
+  const bool mouseSwapped = GetSystemMetrics(SM_SWAPBUTTON);
+  map[Event::LeftButton] = GetAsyncKeyState(mouseSwapped ? VK_RBUTTON : VK_LBUTTON) < 0;
+  map[Event::RightButton] = GetAsyncKeyState(mouseSwapped ? VK_LBUTTON : VK_RBUTTON) < 0;
+  map[Event::MiddleButton] = GetAsyncKeyState(VK_MBUTTON) < 0;
+  map[Event::X1Button] = GetAsyncKeyState(VK_XBUTTON1) < 0;
+  map[Event::X2Button] = GetAsyncKeyState(VK_XBUTTON2) < 0;
+}
+} // namespace
 
 // Converts an os::Hit to a Win32 hit test value
 static int hit2hittest[] = {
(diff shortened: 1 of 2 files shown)


dacap:
I wasn't able to test this yet, but I'm not sure about this solution, probably we could generate a `WindowLeave` and then a `WindowEnter` event? similar to https://github.com/aseprite/aseprite/commit/c31aadae9 I need to give a try on Windows yet to check the bug it fixes first.

ckaiser:
It feels more like something that should be handled at the LAF level transparently, since it's a platform quirk combined with a quirk of how we handle mouse input (storing the keys based on events instead of providing ways to query the mouse), but I'm open to suggestions on a different way to go about it

dacap:
Just testing on Linux. Pressing all mouse buttons (left/middle/right) on the left window, then Alt+tab to the right window, and release those buttons, laf sends a mouse up for each button to the left window and the capture is lost:

https://github.com/user-attachments/assets/76e85e20-0569-48ba-8f78-6f3aa2896242

I'll test this on Windows later, but I think with this PR we can match the Linux behavior.

dacap:
After reviewing the original bug it looks like a specific problem of borderless windows with custom hit areas (title bar, etc), which is the case of Aseprite children windows. And only for the left mouse button (which is the one that triggers the window dragging/resizing). 

I've found that WM_EXITSIZEMOVE/WM_EXITSIZEMOVE is not received if we just click the window title bar (without moving the mouse) and it produces the same issue if we don't generate a MouseUp message from other place.

I've created a new PR https://github.com/aseprite/laf/pull/191 which should be enough but feel free to give a try and test it (anyway I'll merge it ASAP for tomorrow release).

PR #185, opened 2026-05-28
change misc/platform.sh to check if arch command exists

RiverDewberry:
In artixlinux, an arch based distro, `etc/os-release` does not define
`ID_LIKE`. The build script fails as a result of only checking if
``"$ID" == *"arch"* || "$ID_LIKE" == *"arch"*`` because `$cpu` ends up
being unbound, leading to unfound skia binaries.

### DIFF
--- a/misc/platform.sh
+++ b/misc/platform.sh
@@ -14,10 +14,10 @@ if [[ "$uname" =~ "MINGW32" ]] || [[ "$uname" =~ "MINGW64" ]] || [[ "$uname" =~ 
 elif [[ "$uname" == "Linux" ]] ; then
     is_linux=1
     source /etc/os-release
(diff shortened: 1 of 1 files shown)


dacap:
@RiverDewberry is there a way to detect these kind of "arch based" distros? the same bug was reported for CachyOS (#167) but I think `ID_LIKE` should specify "arch" as it's a distro based on arch.

RiverDewberry:
@dacap Since misc/platform.sh seems to only check for arch based distros to determine whether to use `uname` or `arch`, I think (#167) had the correct idea of just using `command -v arch` to check which command to use. I've added a commit to the pr to implement this change.

Also, there a case to be made that artix should not set `ID_LIKE` as it may not have systemd.

dacap:
@RiverDewberry I like this change, I'll run CI and merge it if everything goes ok.

PR #187, opened 2026-05-28
[x11] Add kdialog support, more robust detection

ckaiser:
Adds kdialog support and improves the check (popen was either racing or something else strange was happening because it was always defaulting to the built-in Aseprite dialog except when I was stepping through with the debugger to give it time, std::system will always block).

### DIFF
--- a/dlgs/file_dialog.h
+++ b/dlgs/file_dialog.h
@@ -63,6 +63,8 @@ public:
     // Connection to the X11 server (the Display* pointer returned by
     // XOpenDisplay()).
     void* x11display = nullptr;
+    enum class Backend : uint8_t { Autodetect, KDialog, Zenity };
+    Backend backend = Backend::Autodetect;
 #endif
   };
 
--- a/dlgs/file_dialog_x11.cpp
+++ b/dlgs/file_dialog_x11.cpp
@@ -13,11 +13,11 @@
 #include "base/file_handle.h"
 #include "base/fs.h"
 #include "base/log.h"
-#include "base/replace_string.h"
 #include "base/split_string.h"
 
 #include <X11/Xlib.h>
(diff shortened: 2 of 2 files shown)


dlgs/file_dialog.h
```diff
@@ -63,6 +63,7 @@ class FileDialog : public base::RefCount {
     // Connection to the X11 server (the Display* pointer returned by
     // XOpenDisplay()).
     void* x11display = nullptr;
+    std::string backend = "autodetect";
```
dacap: I'd prefer if this is an enum (probably the CLITool could be moved inside FileDialog), and we convert the string <-> enum from the Aseprite side.

ckaiser: Added the enum and made the backend selection fall back to whatever it finds that is working

dlgs/file_dialog_x11.cpp
```diff
@@ -190,30 +190,107 @@ class FileDialogX11 : public FileDialog {
...
+        std::array<char, 512> buffer;
+        std::string result;
+        while (fgets(buffer.data(), buffer.size(), f) != nullptr)
+          result += buffer.data();
+
+        switch (const int ret = pclose(f)) {
+          case 0:
+            m_filenames.clear();
+            base::split_string(result, m_filenames, "\n");
+            m_filenames.pop_back(); // Superfluous \n
```
dacap, line 252: I think we should check if `!m_filenames.empty()` and `m_filesnames.back().empty()` before doing this `pop_back`.

dacap:
There is just that minor change, but after that I think we can merge it :+1: looks good.

dacap:
Looks great :file_cabinet: :rocket:

PR #188, opened 2026-05-29
Use stable_sort instead of sort

martincapello:
While reviewing https://github.com/aseprite/aseprite/pull/5829 I found the actual fix to aseprite/aseprite#5624.

So today I've learned that std::sort doesn't work the same in all platforms when the elements are equals, this is the reason why in Linux the "packed" sprite sheet presented a different result than in Windows and macOS. So using std::stable_sort makes the relative order of the original collection to stay the same when the elements are equals across all platforms.


dacap:
LGTM :+1:

PR #194, opened 2026-08-11
Fix application slows over time on MacOS (fix aseprite/aseprite#3256)

Gasparoken:
Verified with instruments.

Without the fix: ￼
Screenshot 2026-08-11 at 3 48 32 PM

With the fix:
Screenshot 2026-08-11 at 3 48 15 PM

 ￼
fix [aseprite/aseprite#3256](https://github.com/aseprite/aseprite/issues/3256)


dacap:
LGTM 👍

PR #200, opened 2026-09-02
[x11] Open URLs and files without interpolating into a shell

giobuilds:
Companion to the Aseprite hardening PR (launcher is in this submodule).

`open_url` / `open_file` on Linux and macOS currently run:

    system("setsid xdg-open \"" + file + "\"")

A news RSS or updater `url=` value that contains `"` can therefore become command injection when the user clicks a Home-tab link. This replaces that with `fork`/`execlp`/`waitpid` so the path is an argv element, and rejects non-`http(s)`/`mailto` URLs and control characters.

**Disclosure:** this patch was produced with assistance from a coding agent. Igara's contributing guide says AI-authored code will not be merged. Please close this if that still applies; the change is also on https://github.com/giobuilds/laf/tree/harden for a personal fork.


giobuilds:
Opened against the wrong repository — this was meant for the giobuilds fork only, not upstream.

PR #203, opened 2026-09-22
[win] Sync IME windows while dragging

Cerallin:
Update IME composition/candidate window positions on `WM_MOVING`/`WM_MOVE`, using a cached client-relative caret position to avoid stale screen coordinates.

ime_dragging

I declare that my contributions are not co-authored using a generative AI technology.


Cerallin:
Another approach is to update the candidate window position only when the drag is released. Maybe better?

